PR #1, opened 2015-10-23
0001 - Initial RFC Proposal

daschl:
@couchbaselabs/sdk-team

### DIFF
--- a/review/0001-rfc-process.md
+++ b/review/0001-rfc-process.md
@@ -1,0 +1,90 @@
+# Meta
+
+ - RFC Name: The RFC Process
+ - RFC ID: 0001
+ - Start Date: 2015-10-23
+ - Owner: Michael Nitschinger
+ - Current Status: Review
+
+# Summary
+This RFC describes the newly established RFC process.
+
+# Motivation
+Since the inception of the "SDK", we struggled with inconsistencies between the different language implementations. Many different reasons are contributing to it (legacy codebases, customer escalations, forgetting to communicate,...), but in the end there are two noticable problems to the user:
+
+ - Syntactical API inconsistencies
+ - Semantical behavior differences in cluster communication
+
+Both lead to user dissatisfaction. The first one is often caught during development, the second one very often hits the user in production. In order to provide a better user experience, we should strive to improve our communication and synchronization activities.
+
+Note that this is different from PRDs that PM creates. These RFCs should get into the weeds and nitty gritty details of technical implementation and behavior.
+
+The RFC process is designed to be lightweight, collaborative and effective. Doing this in an open forum also allows easy access and comments by external contributors.
+
+Before creating an RFC, it needs to be clear if an RFC is actually needed. The following describes each situation where an RFC is mandatory:
+
+ - SDK behavior or API changes when interacting with the User OR the Server. So every time either the user or the server communication changes, there is a high likelyhood other SDKs have a similar impact. This explicitly includes utility classes
+ which are very likely to be needed on other SDKs.
+ - The RFC process itself needs to be changed.
+ - Existing inconsistencies that need to be resolved across the board.
+
+If a prototype is already ongoing, it makes sense to drive the RFC alongside the prototype, but this is not a requirement.
+
+Two examples: if the java client includes a helper class for retry semantics on rxjava observables (to reduce the verbosity in writing it manually), this does not have an equivalent on any other SDK and therefore doesn't need a RFC. If the java client introduces a helper class for LDAP support, this does have an impact on the other SDKs and needs a RFC.
+
+When it is not clear if an RFC needs to be created, a quick email or discussion in the team should make it clear wheter the change has an overall impact or not.
+
+**An RFC MUST be signed off by each SDK representative and consensus must be reached. Once the RFC is accepted, each SDK MUST implement it according to the RFC. If, after acceptance issues show up during implementation, the MUST lead to a revision of the RFC which again needs to be signed off by everyone.**
+
+# General Design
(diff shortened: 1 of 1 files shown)


simonbasle:
this is a great idea :smile:
my general thoughts:
 - should we start with a more limited scope, eg what Todd's been calling "verbs" in the `Cluster`|`ClusterManager`|`Bucket`|`BucketManager` API? in current state for instance, it looks like if I want to add a utility class in Java SDK I have to go through a RFC?
 - teams could be described in a file at the root of the repo (possibly in YAML or machine-parsable format, so that in the future a bot could be made to track signoffs and consensus)
 - merge commit message of the PR into master (once accepted) should contain link to the PR and list of signoffs rather than having that in the body of the PR itself (again, easier to automate in the future)

I also have still to decide on how to best archive discussions (esp. could we reach one plateau of consensus/signoff for the general content, then another for the sdk specifics, and how to represent that in the process)...

simonbasle:
here is what I propose as a more detailed process:

- look at latest issue to determine the rfc number (4 digits)
- the `RFCID` is in the form *`rfcNumber`***-***`rfcShortName`*
- open an issue titled `RFCID`, short description of the rfc, does the owner think it requires sdk specifics, etc... tag as appropriate (if we decide to put tags in). the issue allows to "reserve" the rfc number prior to any file being added publicly to the repo.
- create the rfc branch: `drafts/RFCID`
- create the rfc file as `RFCID.md`in `/rfc` directory
- start the discussion by doing a Pull Request to master
- **phase one** iterates on the global idea + "proof of concept" sdk specifics proposed by the owner
- *end of this phase should probably be marked by a first informal round of votes?*
- **phase two** iterates on sdk specifics: each sdk team/representative can then push commits in the branch to improve on his own sdk specifics
- once everyone has contributed sdk specifics, the RFC must be signed off by a message following this convention:
```
rfc is (:+1:|:-1:) for `teamname` *anything after is a vote comment*
```
examples:
> rfc is :-1: for `java` - I'm not happy with the java specifics after toying with it
> rfc is :+1: for `python`

 - once content is :+1: by all teams, merge into master using the following strategy (it retains discussion on intermediate commits, but produces a linear history in master):
  - `git merge --squash -e drafts/RFCID`
  - `git commit`
  - commit message with at least two content rows `Close #XX (see link for discussion)` (to close the pr) and `Close #YY` (to close the issue)

jeffrymorris:
Looks good, but as Simon stated, an RFC should be scoped to a feature, component, etc...something that is going to implemented SDK wide.  A utility class would not fit in this criteria.

brett19:
Hey Guys,
- I think the process described by Simon should work, at least as a first-pass attempt, and we can maybe make some modifications to that process in the future if we see room for improvement.
- I believe an RFC should cover any "external" behaviour of an SDK.  This means anything which affects a user-facing API (not including 'internal' APIs which happen to be public for whatever reason).  This also includes any changes which affect the behaviour of an SDK as it related to communication with an external service (like memcached, n1ql, views, management, etc...).
  - We will need to define some kind of methodology to determine whether a feature should have an RFC, or is SDK specific.  As an example case to think about: The Go SDK implements the GetRandom server API to retrieve a random document.  This is implemented under an 'Internal' API used only by internal Couchbase tools, however, while this API may have made sense to be internal in the beginning, it likely may become a 'Public' API in the future, how do we handle this.
- We should create RFC's which reflect existing SDK behaviour, this includes crud, views, durability, prepared statements, management, etc...  This will provide a path for us to resolve existing inconsistencies between the SDKs that have not been caught so far.
- We need to implement some checks and balances that allow us to verify that an SDK adheres to the RFC as defined, this probably should involve QE.
Cheers, Brett

daschl:
Thanks for all the comments so far @brett19 @simonbasle @jeffrymorris - I'll try to incorporate all of this in a next iteration.

@avsej @mnunberg (+ @ingenthr @ToddGreenstein) can you also please provide your input? Thanks!

daschl:
Okay, updated.

Let's do it the following: once a majority for now agrees on the process specified, I'll close this PR and do it like the way it's supposed to be.

@couchbaselabs/sdk-team please :+1:  or :-1: (this time just for the process itself first, not the full rfc)

:+1:

avsej:
:+1:

simonbasle:
:+1:

Le mer. 11 nov. 2015 08:13, Sergey Avseyev <notifications@github.com> a
écrit :

brett19:
:+1:

mnunberg:
:+1:

ToddGreenstein:
:+1:

jeffrymorris:
:+1:

ingenthr:
Question:
- Frequently discussion of the proposal can be an important artifact, much as the proposal itself.  Does this mean these discussions should happen over github issues?  The goal would typically be that once accepted, one doesn't need to read the discussions, but it's useful to do so when ambiguity or understanding of design decisions comes up.

Comments: 
- I sort of wonder if we need a name other than RFC, since that could be ambiguous with IETF RFCs.  I have no strong opinion on this.
- I would prefer to see a FSM of status listed somewhere near the top rather than phases.  I'd propose: DRAFT, REVIEW, ACCEPTED.
- I do agree with @simonbasle that an expert group of sorts should be created per RFC.  All are welcome to comment and those with strong opinions should be on the EG.

Some of the above are ideas I was pursuing with the onepagers.  Also worth reading: https://www.joyent.com/blog/requests-for-discussion

daschl:
Closing this one for proper "approach". Let's take the further discussion over there (@ingenthr)

PR #2, opened 2015-11-11
0002-subdocapi

mnunberg:


daschl:
@couchbaselabs/sdk-team ping for review

mnunberg:
Pushed a new commit.. this should address the concerns..

mnunberg:
@daschl @brett19 @jeffrymorris @avsej

draft/0002-subdocapi.md
```diff
@@ -0,0 +1,351 @@
...
+Subdocument operations come in two flavors. One is the simple single-operation
+form which takes a single document ID and a single path. Another is a
+_multi-path_ form which retrieves or mutates _several_ paths within a document.
+
+Single-path methods are slightly more efficient and require less processing
+overhead than their multi-path equivalents (that is, even if you pass a single
+path to a multipath method, some additional processing overhead may be involved).
+
+Note that both 'multi-path' and 'single-path' APIs operate on the same
+document, and correspond to a single binary operation. This should not be
```
daschl: can you clarify here in a sentence or two on the impact of multi-path when one path op fails and the other succeeds? what are the expected guarantees?

draft/0002-subdocapi.md
```diff
@@ -0,0 +1,351 @@
...
+regarding whether the value "truly" contains the entire document value, or
+merely a subdoc.
+
+The `MetaDoc` object itself is the return value from normal KV operations
+which contain the value, cas, mutation token, etc.
+
+This object is used as an example, some SDKs may wish to either use the
+'value' or 'document' field of the existing `MetaDoc` return value, or
+subclass it, replacing value with 'fragment'.
+
```
daschl: One comment on naming: I was under the impression that in the SDK 2.0 initiative the doc is called `Document` and its content `content`? So for consistency reasons would make  `DocumentFragment` sense which has the same `content` field? So the user only needs to distinguish between a `Document` and a `DocumentFragment`.

draft/0002-subdocapi.md
```diff
@@ -0,0 +1,351 @@
...
+    `foo.some_array[0]`
+
+or
+
+    `foo.some_array[-1]`
+
+As in N1QL, special path characters may be escaped via a backtick (`` ` ``).
+
+Some operations (notably, those taking a path to an array) may accept the
+_empty path_ which refers to the entire document. This is valid if the document
```
daschl: what is the "empty path"? an empty string being passed in as the path?

draft/0002-subdocapi.md
```diff
@@ -0,0 +1,351 @@
...
+
+Error codes are common to most operations.
+
+### `get_in(docid, path)`
+Gets a single path from the document. Returns a MetaDocFragment containing the
+path value.
+
+* `docid`: Document ID
+* `path`: The sub-document path
+
```
daschl: what should be returned if the fragment is not found? Is there a difference between the docid is not found or the path in the doc is not found?

draft/0002-subdocapi.md
```diff
@@ -0,0 +1,351 @@
...
+of it another way, the path matching algorithm in the server works via simple
+string comparison, so a JSON key `"bs\\ and \t tabs are here" would be matched
+by a path component of `"bs\\ and \t tabs are here"`.
+
+The SDK MUST verify that a path is non-empty for operations which do not
+support empty paths. All other forms of path verification is done on the server
+(but an empty path for unsupported operations will return a memcached `EINVAL`
+which isn't really helpful).
+
+## Single-path Methods
```
daschl: overall note: can you add a intended return value for each of the methods?

draft/0002-subdocapi.md
```diff
@@ -0,0 +1,351 @@
...
+
+### `get_in(docid, path)`
+Gets a single path from the document. Returns a MetaDocFragment containing the
+path value.
+
+* `docid`: Document ID
+* `path`: The sub-document path
+
+### `exists_in(docid, path)`
+
```
daschl: does this return a boolean?

draft/0002-subdocapi.md
```diff
@@ -0,0 +1,351 @@
...
+Like `get_in`, but does not return the value (only checks if it exists)`
+
+### `upsert_in(docid, path, value, create_parents, cas, ttl, persist_to, replicate_to)`
+Unconditionally places the specified value in the provided path.
```
daschl: what errors can be returned here? For example if create_parents is false but the path is not there in the target doc? or the path is invalid?

draft/0002-subdocapi.md
```diff
@@ -0,0 +1,351 @@
...
+  should be created. By default, only the _last_ path element is created if
+  it does not exist.
+* `cas`: Exactly the same meaning as it has in top-level CRUD operations.
+* `ttl`: Expiration time. This has the same meaning as it does in the CRUD
+  operations. The TTL can be set for _all_ mutation methods.
+* `persist_to`, `replicate_to`: Durability requirements. Functions exactly as in CRUD
+
+### `insert_in(docid, path, value, create_parents, cas, ttl, persist_to, replicate_to)`
+Exactly like `upsert_in`, except the path must not already exist.
+If it does, `PathExists` is thrown.
```
daschl: Is there a difference between DocExists and PathExists?

draft/0002-subdocapi.md
```diff
@@ -0,0 +1,351 @@
...
+
+### `replace_in(docid, path, value, cas, ttl, persist_to, replicate_to)`
+Exactly like `upsert_in`, except the penultimate path must exist. If it
```
daschl: same question as above if the doc does not exist?

draft/0002-subdocapi.md
```diff
@@ -0,0 +1,351 @@
...
+### `insert_in(docid, path, value, create_parents, cas, ttl, persist_to, replicate_to)`
+Exactly like `upsert_in`, except the path must not already exist.
+If it does, `PathExists` is thrown.
+
+### `replace_in(docid, path, value, cas, ttl, persist_to, replicate_to)`
+Exactly like `upsert_in`, except the penultimate path must exist. If it 
+not exist, `PathNotFound` is thrown. Unlike `insert_in` and `upsert_in`,
+`replace_in` can work on both dictionary and array values
+
+### `extend_in(docid, path, value_or_values, direction, create_parents, cas, ttl, persist_to, replicate_to)`
```
daschl: any reason we are not doing `append_in` and `prepend_in`?

draft/0002-subdocapi.md
```diff
@@ -0,0 +1,351 @@
...
+
+### `arrayinsert_in(docid, path, value_or_values, cas, ttl, persist_to, replicate_to)`
+Insert an element(s) at a given position in the array. The path should indicate
+the desired position at which to insert the new element(s).
+
+**FIXME:** This should really be `insert_in`, but we are already using that for
+the object-based insert.
+
+### `addunique_in(docid, path, value, create_parents, ttl, persist_to, replicate_to)`
+Adds a 'unique' element into an array (at which end is not defined or specified).
```
daschl: is the reason for this method that exists + extend_in are not atomic together? I'm curious - why does this not have the append/prepend semantics to it (so why is the where part unspecified?)

draft/0002-subdocapi.md
```diff
@@ -0,0 +1,351 @@
...
+### `counter_in(docid, path, delta, create_parents, cas, ttl, persist_to, replicate_to)`
+Performs an atomic counter operation on a given path.
+
+* `delta`: The numeric delta, this is simply added (`+`) to the existing value.
+  This value may be signed to 'decrement' or unsigned to 'increment' the existing
+  value.
+
+See "Numeric Errors" below for a list of errors thrown by this operation.
+This method also returns the current value as a `MetaDocFragment`
+
```
daschl: which errors can happen here if the counter is not there? or doc is not there?
What are the boundaries of delta and is there over/underflow possibility?

draft/0002-subdocapi.md
```diff
@@ -0,0 +1,351 @@
...
+* `MutationSpec.addunique(path, value)`. See `addunique_in`
+* `MutationSpec.counter(path, delta)`. See `counter_in`
+* `MutationSpec.delete(path)`. See `delete_in`
+
+### `lookup_in(docid, lookup_specs...)`
+
+Retrieve (or check the existence of) multiple paths.
+
+Returns a list of MetaDocFragment objects. Individual objects may also
+contain non-success status codes (i.e. some paths may exist while others may fail).
```
daschl: this needs to be further clarified, because right now a Fragment (or document) does not have a status code attached to it?

draft/0002-subdocapi.md
```diff
@@ -0,0 +1,351 @@
...
+
+Since this operation can have mixed results, exceptions should not be raised (or
+if they are, ensure they are extremely well documented).
+
+### `mutate_in(docid, cas, ttl, persist_to, replicate_to, mutation_specs...)`
+Mutate paths in a document. Either all or none of the mutations will succeed.
+
+Upon success, a success return value is provided (i.e. a new `MetaDoc` object).
```
daschl: this will not include the new document, right? just saying "yup its done"

draft/0002-subdocapi.md
```diff
@@ -0,0 +1,351 @@
...
+
+* `delta`: The numeric delta, this is simply added (`+`) to the existing value.
+  This value may be signed to 'decrement' or unsigned to 'increment' the existing
+  value.
+
+See "Numeric Errors" below for a list of errors thrown by this operation.
+This method also returns the current value as a `MetaDocFragment`
+
+### `delete_in(docid, path, cas, ttl, persist_to, replicate_to)`
+Removes a path from the document
```
daschl: I think this needs to be called remove() to be consistent with the bucket-level API `remove`

daschl:
Great second iteration @mnunberg! Aside from the inline comments, I'd like to propose something "out of band" here. After discussing it with @brett19 and @simonbasle a few weeks ago and thinking about it further, here is the thing: I feel if we add all those methods 1:1 this is going (with all the possible overloads on each method) to completely overload the `Bucket` API.

So I sat down and tried to rearrange the API as spec'ed out and came to something which I feel helps with that (and other things). 

So the idea is that instead of putting all methods on the bucket level, we'd only expose two bucket level methods: `lookup_in(docid)` and `mutate_in(docid)`. When you call that method, you get some language-dependent form of a "builder" which allows you to specify something along the lines of:

```
lookup_in(docid)
	-> get(path)
	-> exists(path)

mutate_in(docid, cas, ttl, persist_to, replicate_to)
	-> upsert(path, value, create_parents)
	-> insert(path, value, create_parents)
	-> replace(path, value)
	-> extend(path, value_or_values, direction, create_parents)
	-> array_insert(path, value_or_values)
	-> add_unique(path, value)
	-> counter(path, delta)
	-> remove(path)
```

You can see that you can "chain" individual ops together and in the end it will be executed in the SDK. With this approach, note the following:

 - It reduces the bucket level surface to only 2 methods (or a few more with overloads, but only 2 verbs).
 - It removes the need for the user to "learn" what a multi path spec is. If the user specifies only one call like `->get()` we'll use the single style call underneath, if the user chains up more than one call we transparently will switch over to the multi-path calls.

What do you guys think about that idea?

mnunberg:
You can avoid using the Spec object in Java and perhaps other languages in which the "builder" pattern is idiomatic. Do bear in mind that builders are not idiomatic to all languages.

draft/0002-subdocapi.md
mnunberg: Will do.... though I would like to stress that making any kind of "Fragment" is optional.

draft/0002-subdocapi.md
mnunberg: yes

draft/0002-subdocapi.md
mnunberg: I don't believe the CAS is returned, just as it's not returned for any failed operation.

draft/0002-subdocapi.md
mnunberg: The return values for each of these commands have their analog in the like-named top-level KV methods. So for example, get_in returns the same as get, upsert_in returns the same thing as upsert, etc.

draft/0002-subdocapi.md
mnunberg: I would say no, and that `exists_in` should function like a value-less `get_in`. Less API

draft/0002-subdocapi.md
mnunberg: `PATH_ENOENT`.

draft/0002-subdocapi.md
mnunberg: Just the path/fragment. If the doc doesn't exist, the server will return a `KEY_ENOENT` or similar.

draft/0002-subdocapi.md
mnunberg: same answer!

draft/0002-subdocapi.md
mnunberg: Because those names are taken to mean string append/prepend at the top-level. I think it's important not to conflate the two terms (even if it's unfortunate that `append` has primacy over `append_in`). This is still up for discussion though.

draft/0002-subdocapi.md
mnunberg: Yes, exists and extend_in are not atomic. Additionally, `addunique` ensures the _value_, not the _path_, is unique - and is the only one of the commands which inspects the value. Note that 'uniqueness' can also be enforced by a 0-value dictionary (but that would just end up using more space, and you could only use strings)..

draft/0002-subdocapi.md
mnunberg: This is for the SDK to figure out. In Python for example, the return value is simply a list of (`status, value`) tuples; where a non-zero status indicates something other than success.

Note that the status is not attached to the fragment itself, but to the payload; meaning the `content` (or whatever is being used to contain the "value") is actually a collection of smaller results.

draft/0002-subdocapi.md
mnunberg: right

draft/0002-subdocapi.md
mnunberg: ok

draft/0002-subdocapi.md
```diff
@@ -0,0 +1,351 @@
...
+Note that both 'multi-path' and 'single-path' APIs operate on the same
+document, and correspond to a single binary operation. This should not be
+confused with 'multi' operations in some SDKs which pipeline multiple binary
+requests in the network.
+
+## Objects
+
+MetaDocFragment
+
+This subclass contains the same fields as the MetaDoc, except the `value`
```
jeffrymorris: "the same fields as MetaDoc.." what is MetaDoc? Can you introduce the term before referring to it?

draft/0002-subdocapi.md
jeffrymorris: I tend to agree here with @daschl, `append_in` and `prepend_in` are more natural feeling.

mnunberg: I agree too, but we don't want to confuse terminology. The worry here is that someone will see the definitions of `append_in` and assume that `append` is a similar method, and will get surprised when things don't work. I'd recommend we have a vote on this, I just feel uneasy about the conflict of naming.

draft/0002-subdocapi.md
```diff
@@ -0,0 +1,351 @@
...
+| Generic | .NET         | Java   | NodeJS | Go | C | PHP | Python | Ruby |
+| ------- | ------------ | ------ | ------ | -- | - | --- | ------ | ---- |
+| foo()   | foo<dynamic> | Foo<D> | ...... | .. | . | ... | ...... | .... |
+
+
+## Notes on method naming
+
+Overloading classic KV operations to use an additional _path_ parameter is
+discouraged due to fundamental differences between `get(key)` and `get(key, path)`.
+The latter is not simply a modifier of the former (making the `get` more restricted)
```
jeffrymorris: Replace "restricted" with "specific".

draft/0002-subdocapi.md
```diff
@@ -0,0 +1,351 @@
...
+but rather an entirely different operation.
+
```
jeffrymorris: I disagree here, the action here is the same, the result is slightly different and definitely more specific than the more generic `get`. The internal implementation may be completely different, but from a user's perspective they are just getting a doc or a subdoc from that doc.

mnunberg:
@daschl I would also argue that most of our top-level KV operations don't belong anyway. Considering that a subdoc operation is to a document what a kv operation is to a bucket, I think it's fitting they be placed at the top-level. There are also protocol-level differences between single and multi-ops, and finally the question of whether single ops should even exist considering there is currently little difference (the only effective difference, other than a few extra bytes at the protocol level, between single and multi ops is manifest in `counter_in`, where the multi-variant (i.e. in `mutate_in`) does not return the new current counter value).

draft/0002-subdocapi.md
mnunberg: Considering I don't deal much with languages which contain overloads, I'd say that this should be left to whatever is idiomatic within the language. Do bear in mind however that using overloads would also make the list of overloads for `get` very big!

draft/0002-subdocapi.md
brett19: I agree that having two methods for this might be better, but this operation isn't similar to the memcached 'append' method, so reusing the name might be a mistake here.  What about something like 'push_in' and 'unshift_in'?

mnunberg: Or maybe whatever is idiomatic for append and prepend in the language? `unshift` sounds pretty foreign to anyone not using JavaScript or Perl, for example.

daschl:
@mnunberg okay, given that the multi return values are different I think it will be hard to get it under the same "builder" then, since we can only have one return value. I'll quickly prototype an API and then we have something meatier to look at I guess :)

draft/0002-subdocapi.md
daschl: I'm asking because we already have bucket-level exists method which returns a boolean (based on observe).

draft/0002-subdocapi.md
```diff
@@ -0,0 +1,380 @@
...
+Like `get_in`, but does not return the value (only checks if it exists)`. If the
+item does not exist, the error code is suitably set in the result (or whatever
+the appropriate error delivery mechanism is for the SDK).
+
+### `upsert_in(docid, path, value, create_parents, cas, ttl, persist_to, replicate_to)`
+Unconditionally places the specified value in the provided path.
+
+* `docid`: Document ID
+* `path`: the sub-document path
+* `value`: the sub-document value. This should be anything JSON-serializable.
```
daschl: should this be called `fragment` to mirror the stuff that you get on get_in?

draft/0002-subdocapi.md
```diff
@@ -0,0 +1,380 @@
...
+This class, used in top-level KV operations, is also employed by the sub-document
+API to return values or indicate status.
+
+[The `Document` _may_ be subclassed into `DocumentFragment` to signal that the
+value contained therein is only _part_ of the document (a fragment).
+
+This subclass contains the same fields as the `Document`, except the `value`
+(or `contents`) is replaced by the `fragment` field. This is to avoid confusion
+regarding whether the value "truly" contains the entire document value, or
+merely a subdoc.
```
daschl: what do you think about making the path part of the fragment? since that way every fragment can be "identified" by a combination of its id and its path...

draft/0002-subdocapi.md
```diff
@@ -0,0 +1,380 @@
...
+  JSON values. The list should be of a special type (MultiValue) so the
+  SDK should not serialize the list itself as JSON.
+* `direction`: `FRONT` or `BACK` depending on whether the item is to be
+  added to the beginning or end of the array
+* `create_parents`: Create the enclosing array and any intermediate elements.
+  If this is not set, then the immediate array must exist.
+
+### `arrayinsert_in(docid, path, value_or_values, cas, ttl, persist_to, replicate_to)`
+Insert an element(s) at a given position in the array. The path should indicate
+the desired position at which to insert the new element(s).
```
daschl: no create_parents here?

draft/0002-subdocapi.md
```diff
@@ -0,0 +1,380 @@
...
+  contain an object or value which indicates "Success" (this can be a NULL or
+  empty value to indicate no error)
+* `value`: The value for the path. This is always empty when using `exists`.
+
+
+### `lookup_in(docid, lookup_specs...)`
+
+Retrieve (or check the existence of) multiple paths.
+
+Returns a list of MetaDocFragment objects. Individual objects may also
```
daschl: if we dont put the path into the fragment, wouldn't then then actually be one DocumentFragment with a list of actual fragments inside? Of course if we put the path into the fragment its ambiguous...

I'm asking since most of the stuff is duplicated (like the id(), cas().. and so on)..

draft/0002-subdocapi.md
```diff
@@ -0,0 +1,380 @@
...
+These methods take a document ID and a list of _specifications_ to perform. There are
+only two methods (`mutate_in` and `lookup_in`).
+
+The specification of each method is highly language specific, to use pseudo code, we will
+pretend there are two static helper classes which generate appropriate specs:
+
+### `LookupSpec`
+Creates lookup specifications. There are two 'static methods'
+
+* `LookupSpec.get(docid)` (takes the same arguments as `get_in`
```
daschl: shouldn't docid actually be path?

daschl:
Alright, here is a first prototype of the API as laid out in the spec: https://gist.github.com/daschl/ee5d3231690d33c0cdbe

One thing I found while writing is that it would make sense to put the path into the fragment, since this makes it uniquely identifiable (since we then have the id and path in the fragment). Also, if you load a fragment, modify it and store it later, you don't always need to manually carry the path around (see in the gist above, it would make the code "cleaner").

draft/0002-subdocapi.md
mnunberg: That’s news to me!

If this is the case then it should be idiomatic to the SDK. I consider exists a “valueless get”; in any event I should probably change the description to indicate that exists is successful by lack of an error

daschl: oh? I thought we added it across the board, people requested it and we did it based on observe.. looks like no rfc process back then *sigh* ;)

mnunberg: I requested it, but nobody had actually told me it went in!

mnunberg:
You might want to enumerate what actually goes into the DocumentFragment. Currently it doesn’t look very pretty (but I suppose that’s the only way to deliver the CAS, and would get unruly because of all the possible option combinations..)

draft/0002-subdocapi.md
mnunberg: No, because there is only one case in which create_parents makes sense, and this is when the path is `[0]`, which is equivalent to a prepend anyway.

draft/0002-subdocapi.md
mnunberg: This is an errror, which I believe I fixed? it shouldn’t return a list of fragments

daschl:
@mnunberg woops, added the fragment into the gist.

mnunberg:
So the `Fragment` does make sense for Java, but perhaps not for languages such as Node or Python.. the question is if Java has a discreet `.upsert(key, value, ....)` or if everything is done via a `Document` object.

mnunberg:
Folks, another question.. `append_in` and `prepend_in` actually allow receiving _multiple_ values; we should probably determine how to encapsulate it.. and whether to expose it at all.

draft/0002-subdocapi.md
mnunberg: I'd say that would be good for languages which are mandating a complex object for input

draft/0002-subdocapi.md
mnunberg: I suppose this would make sense if your language is using a `Fragment` class or similar, however the `Fragment` is not part of the mandated return types (it is merely a suggestion for those languages which are using compound inputs)

mnunberg:
@couchbaselabs/sdk-team I think we might be ready to move this into 'DRAFT'. I will attempt to add any unresolved issues into its section at the bottom. I'd suggest to remove any prior comments on the commit itself and address them as line-items in a new comment; once that's done it can be opened for further review.

Issue #6, opened 2015-12-14
0003-indexmanagement


simonbasle:
This RFC will describe an API to ease the management of GSI Indexes through N1QL, using the `BucketManager`.

The aim is to provide simple API methods for the 80% cases and keep the crafting of N1QL requests (eg. "`CREATE INDEX...`" to the minority of cases with specific needs.

This is to be a cross-SDK RFC with maybe 1 method that could be kept as a Java SDK specificity.

PR #7, opened 2015-12-14
0003 Index Management RFC

simonbasle:
@couchbaselabs/sdk-team please start reviewing and looking into SDK specifics.

The current prototype implementation is in Java, see http://review.couchbase.org/#/c/57530/

See associated issue #6

### DIFF
--- a/rfc/0003-indexmanagement.md
+++ b/rfc/0003-indexmanagement.md
@@ -1,0 +1,154 @@
+# Meta
+
+ - RFC Name: Index Management
+ - RFC ID: 0003-indexmanagement
+ - Start Date: 2015-12-14
+ - Owner: Simon Baslé (@simonbasle)
+ - Current Status: Draft
+
+# Summary
+The aim of the "`Index Management`" RFC is to offer a simplified API in the `BucketManager` to list, create and drop indexes, for the majority of use cases. In this sense, it is focused on GSI indexes and assumes a few other givens. For cases that don't fit in this picture, the fallback for the user is to craft appropriate `N1QL` queries (eg. in Java using the `Index` fluent API).
+
+# Motivation
+The goal is to abstract away regularly needed queries relative to indexes and provide a simple API. This way the user can focus on his business related `N1QL` queries, without having to think twice about the boilerplate of creating indexes.
+
+# General Design
+The API will be added to the `BucketManager` class of each SDK. In its blocking form, the API is made up of **6** base methods:
+ - `listIndex`: lists all the indexes relating to the current `Bucket` using the `system:indexes` N1QL view.
+ - `createPrimaryIndex`: create a primary index on the current Bucket.
+ - `createIndex`: create a secondary GSI index on the current Bucket.
+ - `dropPrimaryIndex`: drop the current Bucket's primary index.
+ - `dropIndex`: drop a specific secondary index of the current Bucket.
+ - `buildDeferredIndexes`: trigger the build of indexes that were created in a deferred fashion (see below).
+
+It is open for discussion wether or not this RFC should also include the following methods in a cross-SDK fashion, or if they should be optional:
+ - `watchIndex`: poll the `system:indexes` N1QL view until a given index changes its status to "online" (or a maximum amount of time has passed).
+ - `watchIndexes`: poll the `system:indexes` N1QL view until a given list of index all become "online" (or a maximum amount of time has passed).
+
+
+The underlying design would be to rely on existing `N1QL` verbs and, for the listing and polling part, to use the `system:indexes` keyspace. Each method will internally create and execute a specific N1QL query according to its parameter and the name of the current Bucket.
+
+Each method will offer a bit of tuning with a couple of parameters, so let's detail a more complete signature for each one.
+
+## Listing Indexes
+Signature:
+```java
+List<IndexInfo> listIndexes()
+```
+
+Remarks:
+The method relies on `system:indexes` with the following N1QL statement:
+```sql
+SELECT idx.* FROM system:indexes AS idx WHERE keyspace_id = `BUCKET_NAME_INJECTED_HERE` ORDER BY is_primary DESC, name ASC;
+```
+
+For SDK in language with no dynamic typing, each row will be converted to a `IndexInfo` object with the following attributes:
+
+```java
+String name
+boolean isPrimary
+IndexType type //enum type GSI/VIEW
+String rawType //raw string type in case a new index type is introduced, for forward-compatibility
+String state //eg. "online", "pending"
+String keyspace
+String namespace
+JsonArray indexKey //or equivalent Json representation
+```
+
+## Creating Indexes
+There are two specificities for index creation (both of primary and secondary indexes): the index could already exist and there is the possibility of deferring the build.
+
+In the first specificity, it makes sense to offer two alternatives: either this is considered an error (as N1QL does) or the creation should be purely and simply ignored.
+This choice can be made through a first `boolean` parameter, `ignoreIfExist`. If it is `false`, then a SDK-related exception should be raised (eg. in Java a `CouchbaseException`).
+
+The second specificity can be triggered by another `boolean` parameter, `defer`. If set to true, the N1QL query will use the "with defer" syntax and the index will simply be "pending".
+
+Signatures:
+```java
+boolean createPrimaryIndex(boolean ignoreIfExist, boolean defer)
+
+boolean createIndex(String indexName, List<String> fields, boolean ignoreIfExist, boolean defer)
+```
+
+## Dropping Indexes
+For index removal, the inverse of `ignoreIfExist` must be put in place: N1QL considers it an error to attempt dropping an index that doesn't exist. So the `ignoreIfNotExist` will drive the behavior of the API in such a case: either `false` and a `CouchbaseException` (or equivalent) will be thrown, or `true` and the invocation will do nothing.
+
+Signatures:
+```java
+boolean dropPrimaryIndex(boolean ignoreIfNotExist)
+
+boolean dropIndex(String name, boolean ignoreIfNotExist)
+```
+
+> **Note**: Unlike index creation, defer the drop of an index is not possible
+
+## Building deferred indexes
+Signature:
+```java
+List<String> buildDeferredIndexes()
+```
+
+Remarks:
+The returned list is the list of indexes that were in **`pending`** state at the time of the call.
+
+The method will first list indexes, only keeping the `pending` ones. It then issues the appropriate N1QL statement from the list of pending indexes (here `INDEX_NAME1` and `INDEX_NAME3`):
+
+```sql
+BUILD INDEX ON `BUCKET_NAME_INJECTED_HERE`(`INDEX_NAME1`, `INDEX_NAME3`)
+```
+
+# Watching Indexes
+>**Note**: this part of the API could be considered optional.
+
+Since the process of building a deferred index is asynchronous, the server immediately returns control to the client. It would be useful to have an API to poll the list of indexes until one (or several) indexes are seen as "online". In the blocking API, this would allow to make sure that an index is ready.
+
+Signature:
+```java
+boolean watchIndex(String indexName, long watchTimeout, TimeUnit watchTimeUnit)
+
+List<IndexInfo> watchIndexes(boolean watchPrimary, long watchTimeout, TimeUnit watchTimeUnit, List<String> watchList)
+```
+
+Remarks:
+The implementation would be based on regularly polling the index view (using `listIndexes()` for instance) until a specific index (or a list of indexes) is seen as `online`.
+
+The `watchTimeout` and `watchTimeUnit` are a way to express a maximum **duration** for the poll. This can be replaced by a language-idiomatic way of representing a duration in each SDK.
+
+An index that doesn't exist is ignored, and in such a case `watchIndex` would return `false`.
+
+`watchIndexes` returns the `List<IndexInfo>` of indexes that
+ 1. were already online or...
+ 2. became online during the watch timeframe.
+Thus it can be empty if indexes were all "pending" and failed to be built during the watch timeframe, or if no index in the provided list exists.
+
+# Language Specifics
+## Java
+ - For secondary index creation, since the Java SDK has an `Expression` type, it could accept `List<Object>` as a `fields` parameter. Each element must then either be an `Expression` or a `String` representing the field to be indexed.
+
+ - For convenience, `createIndex` will offer a signature with varargs syntax (`boolean createIndex(String indexName, boolean ignoreIfExist, boolean defer, Object... fields)`).
+
+ - For `watchIndexes`, the polling should be done with a linearly augmenting delay, from 50ms to 1000ms by steps of 500ms.
+
+ - The asynchronous version of `watchIndex` should return `Observable<IndexInfo>`, emitting the IndexInfo of the index once it becames online.
+
+ - Similarly, the async version of `watchIndexes` would return an `Observable<IndexInfo>` with a single emission for each index **that exists** and becomes "online".
+
+## Unresolved SDK specifics
+ * .NET
+ * NodeJS
+ * Go
+ * C
+ * PHP
+ * Python
+ * Ruby
+
+## SDK without specifics (signed off)
+
+# Unresolved Questions
+ * Can the `watchIndex`/`watchIndexes` feature be implemented easily in each SDK and should it be part of the mandatory implementation of this RFC?
+
+# Final signoff
+If signed off, each representative agrees both the API and the behavior will be implemented as specified.
+
+| Language | Representative | Date       |
+| -------- | -------------- | ---------- |
--- a/rfc/0003-indexmanagement.md
+++ b/rfc/0003-indexmanagement.md
@@ -23,20 +23,20 @@ In case new index-related semantics are added in a version of Couchbase Server 4
 The API will be added to the `BucketManager` class of each SDK. In its blocking form, the API is made up of at least **8** base methods:
 
  - `listIndexes`: lists all the indexes relating to the current `Bucket` using the `system:indexes` N1QL view.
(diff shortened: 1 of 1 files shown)


rfc/0003-indexmanagement.md
```diff
@@ -0,0 +1,156 @@
...
+
+# Summary
+The aim of the "`Index Management`" RFC is to offer a simplified API in the `BucketManager` to list, create and drop indexes, for the majority of use cases. In this sense, it is focused on GSI indexes and assumes a few other givens. For cases that don't fit in this picture, the fallback for the user is to craft appropriate `N1QL` queries (eg. in Java using the `Index` fluent API).
+
+# Motivation
+The goal is to abstract away regularly needed queries relative to indexes and provide a simple API. This way the user can focus on his business related `N1QL` queries, without having to think twice about the boilerplate of creating indexes.
+
+# General Design
+The API will be added to the `BucketManager` class of each SDK. In its blocking form, the API is made up of at least **6** base methods:
+ - `listIndex`: lists all the indexes relating to the current `Bucket` using the `system:indexes` N1QL view.
```
daschl: should this be called `listIndexes` ?

rfc/0003-indexmanagement.md
```diff
@@ -0,0 +1,156 @@
...
+ - `createPrimaryIndex`: create a primary index on the current Bucket.
+ - `createIndex`: create a secondary GSI index on the current Bucket.
+ - `dropPrimaryIndex`: drop the current Bucket's primary index.
+ - `dropIndex`: drop a specific secondary index of the current Bucket.
+ - `buildDeferredIndexes`: trigger the build of indexes that were created in a deferred fashion (see below).
+
+Additionally it is open for discussion wether or not this RFC should also include the following ***2*** methods in a cross-SDK fashion, or if they should be optional:
+ - `watchIndex`: poll the `system:indexes` N1QL view until a given index changes its status to "online" (or a maximum amount of time has passed).
+ - `watchIndexes`: poll the `system:indexes` N1QL view until a given list of index all become "online" (or a maximum amount of time has passed).
```
daschl: IMHO this one subsumes the previous one, since you can always pass in just "one" to get the same effect.

rfc/0003-indexmanagement.md
```diff
@@ -0,0 +1,156 @@
...
+
+
+The underlying design would be to rely on existing `N1QL` verbs and, for the listing and polling part, to use the `system:indexes` keyspace. Each method will internally create and execute a specific N1QL query according to its parameter and the name of the current Bucket.
+
+Each method will offer a bit of tuning with a couple of parameters, so let's detail a more complete signature for each one.
+
+## Listing Indexes
+Signature:
+```java
+List<IndexInfo> listIndexes()
```
daschl: if possible, it would be great to use language-independent API here so its not tied to a particular implementation (and not biased).

rfc/0003-indexmanagement.md
```diff
@@ -0,0 +1,156 @@
...
+String keyspace
+String namespace
+JsonArray indexKey //or equivalent Json representation
+```
+
+## Creating Indexes
+There are two specificities for index creation (both of primary and secondary indexes): the index could already exist and there is the possibility of deferring the build.
+
+In the first specificity, it makes sense to offer two alternatives: either this is considered an error (as N1QL does) or the creation should be purely and simply ignored.
+This choice can be made through a first `boolean` parameter, `ignoreIfExist`. If it is `false`, then a SDK-related exception should be raised (eg. in Java a `CouchbaseException`).
```
daschl: we should probably specify the name of the error/exception, ala `IndexAlreadyExistsException`?

rfc/0003-indexmanagement.md
```diff
@@ -0,0 +1,156 @@
...
+
+Signatures:
+```java
+boolean createPrimaryIndex(boolean ignoreIfExist, boolean defer)
+
+boolean createIndex(String indexName, List<String> fields, boolean ignoreIfExist, boolean defer)
+```
+
+## Dropping Indexes
+For index removal, the inverse of `ignoreIfExist` must be put in place: N1QL considers it an error to attempt dropping an index that doesn't exist. So the `ignoreIfNotExist` will drive the behavior of the API in such a case: either `false` and a `CouchbaseException` (or equivalent) will be thrown, or `true` and the invocation will do nothing.
```
daschl: `IndexDoesNotExistException`?

rfc/0003-indexmanagement.md
```diff
@@ -0,0 +1,156 @@
...
+ - RFC ID: 0003-indexmanagement
+ - Start Date: 2015-12-14
+ - Owner: Simon Baslé (@simonbasle)
+ - Current Status: Draft
+
+# Summary
+The aim of the "`Index Management`" RFC is to offer a simplified API in the `BucketManager` to list, create and drop indexes, for the majority of use cases. In this sense, it is focused on GSI indexes and assumes a few other givens. For cases that don't fit in this picture, the fallback for the user is to craft appropriate `N1QL` queries (eg. in Java using the `Index` fluent API).
+
+# Motivation
+The goal is to abstract away regularly needed queries relative to indexes and provide a simple API. This way the user can focus on his business related `N1QL` queries, without having to think twice about the boilerplate of creating indexes.
```
mnunberg: wording: "This way the user can focus on N1QL queries without having to thing twice..."

rfc/0003-indexmanagement.md
jeffrymorris: My vote: IndexExistsException..."Already" is too verbose.

rfc/0003-indexmanagement.md
```diff
@@ -0,0 +1,156 @@
...
+This choice can be made through a first `boolean` parameter, `ignoreIfExist`. If it is `false`, then a SDK-related exception should be raised (eg. in Java a `CouchbaseException`).
+
+The second specificity can be triggered by another `boolean` parameter, `defer`. If set to true, the N1QL query will use the "with defer" syntax and the index will simply be "pending".
+
+Signatures:
+```java
+boolean createPrimaryIndex(boolean ignoreIfExist, boolean defer)
+
+boolean createIndex(String indexName, List<String> fields, boolean ignoreIfExist, boolean defer)
+```
```
jeffrymorris: I think we need an overload here for complex indexes. For example, index hints:

    CREATE INDEX `pv_ndx_02` ON `default`(`userId`,(meta().`id`),`type`) 
    WHERE ((`type` = 'pv') 
    AND ((to_number(`userId`) % 2) = 0)) USING GSI;

Unless of course, if by "fields" you meant "indexes"...

rfc/0003-indexmanagement.md
jeffrymorris: Perhaps IndexNotFoundException or NoIndexDefinedException...naming is hard.

rfc/0003-indexmanagement.md
```diff
@@ -0,0 +1,156 @@
...
+>**Note**: this part of the API could be considered optional.
+
+Since the process of building a deferred index is asynchronous, the server immediately returns control to the client. It would be useful to have an API to poll the list of indexes until one (or several) indexes are seen as "online". In the blocking API, this would allow to make sure that an index is ready.
+
+Signature:
+```java
+boolean watchIndex(String indexName, long watchTimeout, TimeUnit watchTimeUnit)
+
+List<IndexInfo> watchIndexes(List<String> watchList, boolean watchPrimary, long watchTimeout, TimeUnit watchTimeUnit)
+```
```
jeffrymorris: I am wondering if a callback/event driven API might be better here (or as an option) rather than a blocking API might fit this better? I can see in some cases where I would want to block and in others (very log index build times), i wouldn't...I would like to handle an event and then continue code exec from there.

jeffrymorris:
@simonbasle -

Nice work, I like it.

-Jeff

rfc/0003-indexmanagement.md
simonbasle: Java will always have an asynchronous RxJava based equivalent.

rfc/0003-indexmanagement.md
simonbasle: will go for `IndexDoesNotExistException`

rfc/0003-indexmanagement.md
simonbasle: will go for `IndexAlreadyExistException`

rfc/0003-indexmanagement.md
simonbasle: I used the `java` fenced code block so that the result is syntax highlighted and more readable, but I tried to use things that are generic and self-explanatory enough...

rfc/0003-indexmanagement.md
simonbasle: indeed, removed `watchIndex`

rfc/0003-indexmanagement.md
simonbasle: indeed

rfc/0003-indexmanagement.md
simonbasle: fixed earlier

rfc/0003-indexmanagement.md
simonbasle: Here fields would be `userId`, ... (what you have between parenthesis after `default`).

I think of this API as being a convenience for the most basic cases of index building, not a comprehensive coverage of all index building possibilities.

Building indexes with WHERE clause would be out of scope and involve writing a N1QL query instead, with relevant DSL if available. Note that you can still use `watchIndexes` after triggering deferred build of such a complex index manually.

If will add a note about that scope.

simonbasle:
The RFC has been updated, taking previous feedback into account.

mnunberg:
I think `buildIndexes()` is way too high level. What's wrong with `buildIndexes(listIndexes())` ?

simonbasle:
@mnunberg the BUILD instruction in N1QL has a bug where triggering multiple BUILD in parallel can result in a lock :( I think having a single operation that triggers all the pending indexes it can find is less likely to cause this issue than one that takes an index name/list (as users would then be tempted to call the `buildIndexes` after each `createIndex(defer = true)`)...

`watchIndexes` is flexible enough that user can watch a single index or the entire list returned by `buildIndexes` (`watchIndexes(buildIndexes())`)... The single index variant was removed in this iteration to keep the API lean.

ingenthr:
We shouldn't design the API around a bug though, assuming it really is a bug and will be fixed.  That said, if the watchIndexes() is flexible enough, then it sounds good.

simonbasle:
@ingenthr yes I agree, but in the present case the it aligns with the intent of this API to be a simple, very high level one. So I think `buildIndexes()` and `watchIndexes()` meet these goals and buildIndexes() has the added benefit of not encouraging the conditions in which the bug happens.

simonbasle:
I'm thinking this should hit broader review now so unless there is a veto tonight, I'll mark it as REVIEW instead of DRAFT tomorrow morning (UTC+1).

simonbasle:
pinging @jeffrymorris @brett19 @mnunberg @avsej (see above)

ingenthr:
It looks like we may have problems with array indexes.

Note that I opened JCBC-927 based on trying to use this, and it looks like it won't work with array indexes.  From the examples, I tried this:

```
        // Create an index list like this:
        //   CREATE INDEX isched
        //     ON `travel-sample` ( DISTINCT ARRAY v FOR v IN schedule END );
        // …per the example at http://developer.couchbase.com/documentation/server/4.5-dp/indexing-arrays.html
        String indexString = "( DISTINCT ARRAY v FOR v IN schedule END );";
        boolean createSuccess = bucketManager.createIndex("isched", true, true, indexString);

```

… but reading details behind it seems this kind of string will be escaped and then won't work.

mnunberg:
I think probably safest and future-proof (albeit not the nicest) way to do this is to have an option that omits the insertion of backticks. This might also be helpful if the name already has backticks from elsewhere.

simonbasle:
@mnunberg @ingenthr The idea for index management API in my mind was to cover the 80% use case where you want to build indexes that are fairly simple without the overhead of writing and executing the N1QL request yourself. If you feel that API should go beyond that, or that array indexes or conditional indexes (as in having a where clause) are part of the 80% use case, then this needs further work indeed.

mnunberg:
(i also think the 80% use case doesn't involve async index building either....)

simonbasle:
@mnunberg @ingenthr added two commits: one is about the corner(ish) case of named primary indexes (that should be creatable and droppable with the API) and the other is mentioning a possible way (at least for the Java SDK) around the problem of complex index expressions (like `DISTINCT ARRAY v FOR v IN schedule END`)

mnunberg:
I think it makes much more sense for SDKs like C and Python to have the _watch_ functionality embedded into the _build_ functionality, so that it is included as a subset of its options rather than a different op.

simonbasle:
@couchbaselabs/sdk-team last ping in order to get the expected API minimal example in your respective SDKs + signoff...

As discussed during this week's meeting, this should now move on. If you don't explicitly provide a +1/-1 comment by the end of this week (so 1 day left for North America), I'll consider you accept to follow the RFC as it is.

I will promote this to REVIEW by Monday morning (CEST) for a week of wider-audience review/discussion, then ACCEPTED and merge by Friday April 1st CEST.

rfc/0003-indexmanagement.md
brett19: IndexAlreadyExistsException?

rfc/0003-indexmanagement.md
```diff
@@ -0,0 +1,201 @@
...
+
+# Watching Indexes
+>**Note**: this part of the API could be considered optional.
+
+Since the process of building a deferred index is asynchronous, the server immediately returns control to the client. It would be useful to have an API to poll the list of indexes until one (or several) indexes are seen as "online". In the blocking API, this would allow to make sure that an index is ready.
+
+Signature:
+
+```java
+List<IndexInfo> watchIndexes(List<String> watchList, boolean watchPrimary, long watchTimeout, TimeUnit watchTimeUnit)
```
brett19: There should not be a need for 'watchPrimary' here.  watchIndices should implicitly wait for the primary index to come online as indexing does not function without it, additionally there would never be a case to watch no indices.

simonbasle: mmh what if you didn't defer-build the primary index? (yeah I know, that would be dumb)
actually, if the user takes the output from `buildDeferredIndexes` and feeds it to this, if the primary was deferred it will be included already. This is just meant as a convenience for the case where the user compiles the list by other means and in the present case don't need to know that in N1QL a primary index is named `#primary` by default...

Users might also have a need to watch a particular index only, without being interested in the state of other pending indexes.

Finally, I would have to double check that but doesn't covering indexes bring the possibility that a N1QL query can be executed without any primary index in place? Are you sure your the "indexing does not function without [the primary index]" statement?

brett19: I am relatively sure, but it's certainly something we can bring up with the query team.  In regards to your first statement, even if they didn't defer-build the primary index, it would still need to be built before secondary indexes could function, so there is no downside to ensuring it is built (if its not deferred, its going to be in the process of building by default anyways).

rfc/0003-indexmanagement.md
```diff
@@ -0,0 +1,201 @@
...
+//make sure "byDescAndToto" will be recreated from scratch by dropping it if it exists
+bucketManager.dropIndex("byDescAndToto", true);
+
+//defer creation of the secondary index on field toto and desc.
+//note that DESC is a keyword, but the field is escaped when provided as a String.
+bucketManager.createIndex("byDescAndToto", false, true, x("toto"), "desc");
+
+//since one of the indexes was built with defer, trigger the build and wait for it to finish in maximum 1 minute
+bucketManager.watchIndexes(bucketManager.buildDeferredIndexes(), 1, TimeUnit.MINUTES);
+
```
jeffrymorris: I am wondering if watchIndexes is better represented as an asyn API where a callback is invoked when its completed?

rfc/0003-indexmanagement.md
```diff
@@ -0,0 +1,201 @@
...
+This choice can be made through a first `boolean` parameter, `ignoreIfExist`. If it is `false`, then a SDK-related exception, `IndexAlreadyExistsException`, should be raised for SDKs that rely on exceptions.
+
+The second specificity can be triggered by another `boolean` parameter, `defer`. If set to true, the N1QL query will use the "with defer" syntax and the index will simply be "pending" (prior to 4.5) or "deferred" (at and after 4.5, see [MB-14679](https://issues.couchbase.com/browse/MB-14679)).
+
+Signatures:
+
+```java
+boolean createPrimaryIndex(boolean ignoreIfExist, boolean defer)
+
+boolean createNamedPrimaryIndex(String customName, boolean ignoreIfExist, boolean defer)
```
jeffrymorris: This method would probably be fine as just an overload of createPrimaryIndex.

rfc/0003-indexmanagement.md
simonbasle: I'll edit one last time to clarify that `watchIndexes` is indeed optional and to be implemented in the manner that seems the most idiomatic to each particular SDK. In Java, since we have clearly defined sync/async APIs which mirror each other, this translates to a `List<IndexInfo> watchIndexes` method in the blocking API mirrored by a `Observable<IndexInfo> watchIndexes` in the async API.

rfc/0003-indexmanagement.md
simonbasle: just did a test, not creating a PRIMARY index but instead `CREATE INDEX foo ON default(type)` allows the following statement to work **without the primary index**: `SELECT * FROM default WHERE type = "something"`.

rfc/0003-indexmanagement.md
simonbasle: good point, and same for dropNamedPrimaryIndex...

rfc/0003-indexmanagement.md
ingenthr: I don't think watchIndexes should be optional, but the method of async should be.

Issue #3, opened 2015-11-12
0001-rfc-process


daschl:
This RFC describes the newly established RFC process.

(how meta!)

PR #9, opened 2015-12-16
0004-at_plus

daschl:


daschl:
Pinging @couchbaselabs/sdk-team for review. Sample implementation for java can be found here: http://review.couchbase.org/#/c/57844/

brett19:
Is there no chance that we could hide the details of the MutationTokens in the majority of cases here?  I suspect in most instances, the user will be using MutationToken's from the most recently performed operation anyways.

daschl:
@brett19 do you want to hide the tokens completely or just their internals? The latter part is defined there right now, since the user just takes the opaque mutation token from a mutation result and then passes it into the n1ql query, not concerning about whats in there (thats just in the rfc for implementation detail purposes)

mnunberg:
Some points of the existing implementation in Python

1. There is no need to explicitly define `consistency` here. Adding any kind of mutation token should implicitly convert the `consistency` to `at_plus`. Maybe declare it an _error_ to set `consistency` when mutation tokens are present, or conversely to add mutation tokens when consistency is something other than null.
2. We don't expose the `MutationToken` at all. Rather the document result itself is passed to the n1ql function. The document contains a private field (`_mutinfo`) which contains the mutation token
3. As part of supporting durability, libcouchbase actually contains an array of vBucket -> {uuid,seqno} information. As such the Python SDK exposes a `N1QLQuery.consistent_with_all(bucket)` method which will add all mutation tokens.

See https://github.com/couchbase/couchbase-python-client/blob/master/couchbase/n1ql.py

brett19:
Hey @daschl,
What Mark has defined as point #3 above is more or less what I was thinking, and would cover nearly every real-world use case that I can think of.

daschl:
After thinking about (and discussing) it more, I'm still torn on abstracting it completely away from the user.  I like the idea from an API perspective, but I was thinking about a little differently than what @mnunberg proposed.

For example, we can add a a `LOCAL_PLUS` enum variant which will then internally indicate that we need to grab the tokens we've been storing in the background. That way we still keep the enum semantics and don't need to add anything else to the query options.

That said, here are the potential difficulties with this approach:

 1. We'll start out with a sparse 1024 element array of tokens, but as the application continues to work eventually of those will fill up. Which means that we are going to transfer the seqnos+uuid for all vbuckets on every at_plus request. I did a quick math and doing so has an overhead of around 45k, which is quite a lot. 
 2. By abstracting the explicit tokens away, it is not possible anymore for the user to perform thread-like RYOW isolation, since there is no way to control what other threads are doing (only if the user adds those sync primitives on his own). One thought that plays into this is since it's AT_PLUS as a minimum and not exact bound, you can't be 100% sure anyways that a different player overrode the stuff you are looking for with your seqno. But there is still the case where lets say t1 modifies x1 once but t2 modifies x2 (something else) many times in the meantime, once the query happens for x1 to be visible the seqnos will be much higher, potentially t1 waiting longer than actually needed in the first place.
 3. By abstracting the tokens away there is no way for the user to share them with other app servers or so, but I don't know if that is a real use case. but we talked about it in the past so I'm putting it here.

So in that list only number 1 is a real "issue" I think, while number 2 being an edge case we should think about.

In addition to that, something we need to think about is the ambiguity right now on what happens with at_plus in general on multi bucket queries. This has not been talked through at all and it does not affect the user API, but definitely the final implementation at some level.

daschl:
@couchbaselabs/sdk-team requesting input!

mnunberg:
All of Michael's points are perfectly valid. To answer each point, however:

1. I am not sure why a user would _want_ to make things consistent with 'everything' - although the mere concept itself implies some added overhead which should perhaps be documented. Another option is simply to do `consistent_with_key` which would calculate the vBucket for a given key without requiring the user to keep track of the result object.

2. Valid, but only applicable to threaded clients (conversely, the API in Python is only valid for non-threaded clients)

3. This would be something done at a later point in time.

The real reason behind `consistent_with_all` or `consistent_with_keys` is to specify an API-level middle-ground between `request_plus` (which grabs all the latest information) and an explicit `at_plus` which requires the user to somehow keep track of this extra data in between queries, or to perform this logic themselves. There is of course, API-wise, nothing preventing the user from doing this global map themselves (assuming of course that he's able to get his hands on the vbucket map).

Going back to my original concern about the confusion behind exposing the `MutationToken` in the first place, perhaps that might not be that big a deal. Python has been exposing the `flags` for `get()` responses since the beginning and I don't believe I've ever encountered a question or issue about it; so I'm not _strongly_ opposed to exposing it

jeffrymorris:
@daschl - For multi-threaded environments why not scope the tokens to the request class instead of the bucket? Wouldn't that resolve most issues?

In general I am tending to the more abstracted proposal by @mnunberg and @brett19 for a couple of reasons:

- Smaller API footprint is easier to grok and maintain 
- Internal implementations can change without concern for backward compatibility (assuming the behavior doesn't change as well)

IIRC, @brett19 and @mnunberg are favoring the originally discussed API (from a few months back), I am curious what is motivating the changes proposed by @daschl? Lessons learnt? Platform specific implementation details? I am not being critical here, i am just trying to get a better mental image of how we arrived here. FWIW, the implementations for single threaded and multi-threaded clients might have to deviate a bit if it makes sense to do so.

-Jeff

-Jeff

daschl:
@jeffrymorris 

> IIRC, @brett19 and @mnunberg are favoring the originally discussed API (from a few months back), I am curious what is motivating the changes proposed by @daschl? 

As far as I remember, this was never spec'ed out at all. The closest I remember is [RYOW Consistency Semantics](https://docs.google.com/document/d/1cO_l3cDlWkqh5psoP6K-5vn-16ikagY4-eMYKXLqABA/edit#) by @mnunberg but this does neither include `consistent_with_all` nor `consistent_with_key`.

@mnunberg 

> Going back to my original concern about the confusion behind exposing the MutationToken in the first place, perhaps that might not be that big a deal. Python has been exposing the flags for get() responses since the beginning and I don't believe I've ever encountered a question or issue about it; so I'm not strongly opposed to exposing it

What about the following 

 - if the user just specifies `SCAN_CONSISTENCY#LOCAL_PLUS`, then we go and grab all the tokens. This is the easiest, but it also has some runtime overhead.
 - We have an optional query param which is called `consistent_with[_X]*()` where in langs that allow overloading would be called  with either Strings (that is allowing to specify custom ids) or with mutation tokens. In languages where method overloading is not available you'd have 2 methods: `consistent_with_ids` and `consistent_with_tokens`. If Ids are passed in we grab the right vbuckets, and if the tokens are passed in we use them right away.

mnunberg:
I like the suggestion, other than the `LOCAL_PLUS` which is a bit confusing (and redundant at the protocol layer as well). This also makes several other assumptions about query processing phases which would be difficult to implement (namely that the SDK core has access to the 'proper' JSON request body which isn't really the case in libcouchbase).

But I do think that the various `consistent_with_XXX` (or overload equivalents) are the way to go:

* `consistent_with_tokens(tokens...)`, `consistentWith(MutationToken...)`: Raw token input. Mandatory
* `consistent_with_all(bucket)`, `consistentWith(Bucket)`: Consistent with all previous mutations performed by bucket. Optional
* `consistent_with_keys(keys...)`, `consistentWith(String...)`: Subset of `consistent_with_all`, filtering on vBuckets. Optional

```
rv = cb.upsert('foo', 'bar')
q = N1QLQuery("select....")
q.consistent_with_tokens(rv.mutation_token)
q.consistent_with_keys('foo')
q.consistent_with_all(cb)
```

With respect to `consistency_for_all`, there is also the option of something like this:

```
q = N1QLQuery("SELECT ...")
q.consistent_with_tokens(cb.recorded_tokens())
```

The second form might somehow 'intuitively' bear weight on the performance implications of sending all tokens? :)

daschl:
Great thoughts @mnunberg! Thinking about it, we can really get rid off LOCAL_PLUS if the explicit methods are used. I guess this needs to be a runtime error if a scan consistency is set and some of the other options listed in this discussion.

I actually like the idea of using `consistent_with_tokens` to subsume the functionality of `consistent_with_all`, also for the reason that for example in java it looks weird if you have to pass in the bucket reference if that is the "scope" you are working with. Since you will pass the n1ql query into the bucket instance anyways...

One "problem" I can see with the last approach is that some clever user might cache the `recorded_tokens()` in a variable and then very subtly the code might work in testing but will produce weird failures in production since the tokens get out of sync. In light of this, let me propose the following slight variation of what you proposed:

 - As you said, get rid of AT_PLUS and keep NOT_BOUNDED as well as REQUEST_PLUS and have it fail at runtime if one of them is provided AND one of the options below.
 -  As you said: `consistent_with_keys(keys...)`, `consistentWith(String...)`: filtering on vBuckets.
 - As you said: `consistent_with_tokens(tokens...)`, `consistentWith(MutationToken...)`: Raw token input.
 - No `consistent_with_all`, but rather if `consistent_with_tokens` is empty it is assumed to take all vbucket tokens.

So basically from a user perspective, the simplest approach would be just to use `consistent_with_tokens()` which will  grab all the tokens for the bucket. If the user wants to narrow it down further for performance (or other) reasons he can either do it by explicitly pass in token(s) into `consistent_with_tokens(Tokens...)` or use `consistent_with_keys(keys...)`.

brett19:
I don't think we should overload 'consistent_with_tokens' as 'consistent_with_all', because I think logically, if your being 'consistent with 0 tokens', its the same as NOT_BOUNDED rather than being completely bounded?

daschl:
@brett19  I see - my thought was that it would imply "consistent with the current tokens". To me it seems redundant to pass in a bucket reference you want to be consistent with when you take that query and then pass it into a bucket scope again - but on the other hands its not too bad if this helps us cover all cases...

Let's discuss those final pieces tonight in the meeting!

daschl:
@couchbaselabs/sdk-team updated with the latest bits, let's do a final round of review! Please +1 or -1 with your comments :)

@mnunberg @brett19 I tried to incorporate the last concerns into the latest update, let me know if that works now.

mnunberg:
+2

brett19:
What happens when someone passes a BucketInstance that isn't attached to the bucket your querying, or alternatively, passes tokens from another bucket?  Relatedly, how do we handle cross-bucket queries with this.

daschl:
@brett19 I made two more modifications (based on our discussions last week):

 - The token now also contains the bucket name / a reference to the bucket. this allows the client to check if the passed in token actually conforms to the query the bucket is scoped in.
 - for cross bucket queries, the server team will add the bucket scoping. I already added this scoping into the RFC, but it is subject to change based on the actual API they'll propose. But I suspect it will be something like this

daschl:
@couchbaselabs/sdk-team can everyone please supply their protoype APIs so I can put it in?

Also, I made some more changes (one that states the SDK needs to perform deduplication on mutation tokens), and there are 3 more open questions that came to my mind - ideas?

 - `consistent_with(ids: String...)` is ambiguous in cross-bucket scenarios. Behaviour? 
 - `consistent_with(ids: String...)` is used and some or all of the tokens are not there in the mutation map yet. Behaviour?
 - `consistent_with(buckets: Bucket...)` is used and some or all of the tokens are not there in the mutation map yet. Behaviour?

mnunberg:
It would be impractical for libcouchbase to include a bucket string or reference with each returned mutation token.

It may be possible to do this with Python, though.

daschl:
@mnunberg well I guess the upper level impl needs to keep it somehow - why is it impractical? otherwise you have no way to distinguish a token if it belongs to bucket A or bucket B once you are out of the scope?

rfc/0004-at_plus.md
```diff
@@ -0,0 +1,224 @@
...
+# Language Specifics
+
+## Setting Scan Consistency With Document IDs
+
+ - **Generic:** `consistent_with(ids: String...)`
+ - **C:** ``
+ - **Go:** ``
+ - **Java:** `N1qlParams consistentWith(String... ids)`
+ - **.NET:** ``
+ - **NodeJS:** ``
```
brett19: `consistent_with({ids: String...})`

rfc/0004-at_plus.md
```diff
@@ -0,0 +1,224 @@
...
+| `SCAN_CONSISTENCY` specified in addition to `consistent_with` | IllegalArgumentException or equivalent thrown at runtime |
+| More than one `consistent_with` variant used at the same time | IllegalArgumentException or equivalent thrown at runtime |
+
+# Language Specifics
+
+## Setting Scan Consistency With Document IDs
+
+ - **Generic:** `consistent_with(ids: String...)`
+ - **C:** ``
+ - **Go:** ``
```
brett19: func (q *N1qlQuery) consistent_with_ids([]string) *N1qlQuery

rfc/0004-at_plus.md
```diff
@@ -0,0 +1,224 @@
...
+ - **NodeJS:** ``
+ - **PHP:** ``
+ - **Python:** ``
+ - **Ruby:** ``
+
+## Setting Scan Consistency With Mutation Tokens
+
+- **Generic:** `consistent_with(tokens: MutationToken...)`
+- **C:** ``
+- **Go:** ``
```
brett19: `func (q *N1qlQuery) consistent_with_tokens([]gocbcore.MutationToken) *N1qlQuery`

rfc/0004-at_plus.md
```diff
@@ -0,0 +1,224 @@
...
+- **Java:** `N1qlParams consistentWith(MutationToken... mutationTokens)`
+- **.NET:** ``
+- **NodeJS:** ``
```
brett19: consistent_with({tokens: MutationToken...})

rfc/0004-at_plus.md
```diff
@@ -0,0 +1,224 @@
...
+- **PHP:** ``
+- **Python:** ``
+- **Ruby:** ``
+
+## Setting Scan Consistency With Bucket Reference
+
+- **Generic:** `consistent_with(buckets: Bucket...)`
+- **C:** ``
+- **Go:** ``
```
brett19: `func (q *N1qlQuery) consistent_with_buckets([]Bucket) *N1qlQuery`

rfc/0004-at_plus.md
```diff
@@ -0,0 +1,224 @@
...
+- **Java:** `N1qlParams consistentWith(Bucket... buckets)`
+- **.NET:** ``
+- **NodeJS:** ``
```
brett19: consistent_with({buckets: Bucket...})

rfc/0004-at_plus.md
```diff
@@ -0,0 +1,224 @@
...
+- **PHP:** ``
+- **Python:** ``
+- **Ruby:** ``
+
+## Setting Scan Consistency With Bucket Reference
+
+- **Generic:** `consistent_with(buckets: Bucket...)`
+- **C:** ``
```
mnunberg: `lcb_n1p_mutation_token(lcb_N1QLPARAMS *obj, const char *keyspace, const lcb_MUTATION_TOKEN *mt);`

daschl:
Updated again with all the new APIs I got and putting in the edge cases with proposed solutions I was concerned about before.

Missing before we can get this into *REVIEW*:
 - Missing APIs for .net, python, php and C (and maybe ruby)
 - Finalized query API on how we need to send it over the wire
 - signoff by stakeholders

mnunberg:
For C

```c
lcb_n1p_setconsistent_token(lcb_N1QLPARAMS*, const char *bucket, const lcb_MUTATION_TOKEN *mt);
```

For Python

```python
params.consistent_with_tokens(rv.mutation_token)
params.consistent_with_tokens(bucket.mutation_tokens())
params.consistent_with_tokens(bucket.mutation_tokens(key))
```

daschl:
@mnunberg I'm not sure I understand - how does python map to the three different methods? are really all three `consistent_with_tokens` and not just `consistent_with`?

Also, for C, the spec is missing the one for document ids and the one for mutation tokens - you only sent over one (does the other not make sense?)

daschl:
Okay, only PHP missing (and optionally ruby, but not mandatory). server API is in.

Please all do a final round of :+1:  or comments and once PHP is available I'll move it into review state.

:+1: java

daschl:
@couchbaselabs/sdk-team ^^

mnunberg:
(:+1)

simonbasle:
minor comment (see inline), otherwise :+1:

daschl:
Okay, updated one more time. PHP is not included for now, as is Ruby. 

Which means that the RFC would be ready to :+1: by everyone now if it makes sense, but I'd like to throw one last thing in here that came up when discussing with Brett.

- What we could do to simplify the API is to instead of taking a `String ids...` and a `MutationToken tokens...` is to just take the `Document` which has both the id and the token (optionally) attached to it. This would reduce the API to two methods:

```
consistent_with(documents: Document...)
consistent_with(buckets: Bucket...)
```

The `consistent_with(buckets: Bucket...)` would be the same, and the behaviour for `consistent_with(documents: Document...)` would be as follows:

 - If the token is set on the `Document`, then it is used, together with `AT_PLUS`
 - If the token is not set on the `Document` but mutation tokens are enabled, then it is looked up in the map and if found used with `AT_PLUS`.
 - If the token is not found and mutation tokens are not enabled, or it is looked up and not found, `REQUEST_PLUS` is used instead.

mnunberg:
I don't think we should implement any of this magic behavior. It may potentially lead to hard-to-debug performance issues later on. I see this as a low-level API and should have DWIS (Do what I say) rather than DWIM (Do what I mean) semantics.

consistent_with_docs would be nice. The benefit of just accepting a mutation token is that only one top-level method is required...

To be fair, I don't feel very strongly against method naming (whether we just pass in the docs, have discreet methods etc. etc. -- I think my wavering between all of the options is evidence of that), but I would be against performing such "magic" with a fairly low-level API.

simonbasle:
Personally I like the latest proposal but to work (eg. in java) users have to capture the return value of the mutations *which I'm not sure they all think of doing*... `consistent_with_ids` has the benefit of always working. However it could be the occasion to show it is best practice to use the `Document` returned by a mutation (updated cas, mutation token, better consistency in N1QL queries...).

I'm not too strongly feeling on either side, and the proposal as it stands looks good to me as well.

daschl:
Alright, I can see that the support for it is not enthusiastic - I'm good with that.

So can @couchbaselabs/sdk-team please :+1:  or raise final objections before we move it into review state? Let's plan for EOD 2016-01-27 to make progress.

daschl:
@simonbasle actually, we can easily make this work since you can create a document from just the id, without the content. So one could use `consistent_with(JsonDocument.create("foo"))`.

brett19:
After talking a lot with @daschl, I think we may have come up with an API that solves for all of our 'edge' cases, and additionally allows us to continue to 'hide' MutationToken's from being directly visible to users, at least until we are able to expose them in a more reliable manner (allow consistent serialization and bundling of the tokens).  The proposed API would look something like this:
consistent_with(Bucket)
as well as having an overloaded method:
consistent_with(Bucket, []Document)
This would of course also tie in the with 'fallback' logic which Michael suggested above, although there would be a little less 'magic' happening due to the more strict semantics associated with this.  Note that you could also call consistent_with multiple times now, each time passing a different bucket and different set of documents, if you need to provide consistency across multiple buckets.

daschl:
@brett19 I'll update the proposal with the latest approach, I really think that makes sense. Once updated we can do a final round of comments before getting it into review. Lets pick a hard date Thursday, 11th Feb so there is a "last chance" for everyone to raise objections in our weekly meeting.

daschl:
@couchbaselabs/sdk-team please another round of review and if :+1: then also please provide the API for it!

rfc/0004-at_plus.md
```diff
@@ -0,0 +1,207 @@
...
+
+This method must have the following behavior:
+
+ - When called with just the `Bucket` reference and no `Documents`, this means that
+   the N1QL query must be consistent with all mutations that happened against this
+   `Bucket`.
+ - When called with `Bucket` and a list of `Documents` and the `MutationTokens` are
+   present on the `Document`, then they are extracted and used as part of the query.
+   This is the most performant option in terms of network overhead.
+ - When When called with `Bucket` and a list of `Documents` and the `MutationTokens` is
```
jeffrymorris: Duplicate "when".

rfc/0004-at_plus.md
```diff
@@ -0,0 +1,207 @@
...
+   for that vbucket is used.
+ - Subsequent calls to either a previously captured `Document` on the same `Bucket` 
+   reference will override the old one, there must not be duplicates in the performed
+   query against N1QL.
+
+Setting `SCAN_CONSISTENCY` to anything but the default is considered a runtime
+error (see errors). This means that `AT_PLUS`, while needed internally, is never
+actually exposed to the user level. The user only specifies the consistentWith
+boundaries and the SDK needs to figure out the rest (see implementation).
+
```
jeffrymorris: This is rather inconsistent with how the other consistency settings are handled, but i guess it doesn't make a lot of sense to require a user to call consistent_with(...) and setConsistency(requestConsistency.At_PLUS).

rfc/0004-at_plus.md
```diff
@@ -0,0 +1,207 @@
...
+- If only the `Bucket` reference is passed in, all mutation tokens for this bucket
+  (so up to 1024) need to be selected and marshalled into the query as shown
+  above. This has the highest overhead over the network.
+
+For both of these approaches to work, the SDK needs to maintain a `Token Map` for
+each open Bucket which maps a vbucket to a MutationToken. To keep this map in sync,
+after every mutation and if tokens are enabled, the map needs to be updated with
+the latest token for the touched partitions. To rephrase this slightly: the SDK
+needs to keep the state of all the mutations passed through it in the form of
+Mutation Tokens per vbucket (partition).
```
jeffrymorris: Isn't there potential for the `Token Map` to get extremely large horizontally, for example:

vbid1 : {t1, t2, ... tnth}
vbid2 : {t1, td, ...tnth}
.
.
vbidtnth

Is there any pruning or invalidation that would happen so we don't create our own memory leak here?

rfc/0004-at_plus.md
```diff
@@ -0,0 +1,207 @@
...
+
+# Language Specifics
+
+## Setting Scan Consistency 
+
+ - **Generic:** `consistent_with(bucket: Bucket, [docs: Document...])`
+ - **C:** ``
+ - **Go:** ``
+ - **Java:** `N1qlParams consistentWith(Bucket bucket, Document... docs)`
+ - **.NET:** ``
```
jeffrymorris: IQueryRequest ConsistentWith<T>(IBucket bucket, params IDocument<T>[] documents)

brett19:
DocumentFragment puts a bit of a wrench in these plans.

rfc/0004-at_plus.md
```diff
@@ -0,0 +1,207 @@
...
+    [required] sequence_number = long
+    [required] bucket = string/reference (either name of the bucket or immutable "reference" to the target bucket)
+}
+```
+
+The N1QL query execution for AT_PLUS consists of the following user level API:
+
+```
+consistent_with(bucket: Bucket, [docs: Document...])
+```
```
jeffrymorris: In the .NET client the object responsible for executing the N1QL is the bucket itself, does it makes sense to provide overloads that don't include the bucket itself? Otherwise consistent_with should be defined at the cluster object level (where query _should_ be) as well.

jeffrymorris: Per my comment above, the RFC doesn't seem to define a location for the consistent_with method wrt to object model defined in the SDK 2.0 spec-a-thing. I think needs to be in here to remove ambiguity and/or provide a reference for implementations (and caveats if they can be justified).

jeffrymorris:
@brett19 perhaps DocumentFragment should be either considered and a section added  or even flagged as not supported (in the short term?). In either case it should be addressed before the RFC is "accepted".

mnunberg:
I don't see how `DocumentFragment` affects anything. It's really just a `Document` with a different payload style.

rfc/0004-at_plus.md
daschl: @jeffrymorris yes, the idea was that setting something on `consistent_with` implies "at_plus", so there is no need to set the param explicitly (which means we also need to export it on the enum).

rfc/0004-at_plus.md
daschl: I'll clarify in the spec too, but you shouldn't have more than one token per vbucket, so when a new mutation arrives with a higher seqno for that vbucket you just replace it. So there are always maximum <open buckets> * <vbuckets> stored on the system.

rfc/0004-at_plus.md
daschl: Ah, I guess we need to clarify in the spec, this is on the part where all the other query options are, associated with each query.

Actually I was thinking about this too, if no bucket is provided then the one in scope can be used. The only downside I see with this is that in a cluster-level setting it would need to throw a runtime exception saying it is ambiguous.

daschl:
@mnunberg @brett19 I don't see where it throws a wrench in, but what we can do is make it accept `DocumentFragments` in addition to `Documents`, since all we need is the id and a token, which we have available on the fragment.

daschl:
Okay, I updated the spec with some additional clarifications that hopefully address @jeffrymorris comments.

Now another issue was brought up and that is the subdoc support including the `DocumentFragment`. Right now it is not specified if the `DocumentFragment` exposes its `id()`, which has some implications on AT_PLUS support (since accessing the id is part of the fallback strategy if no token is available). 

Hence I propose the following options which we need to come to an agreement on before moving forward:

- We create a marker interface (for example `MutatedDocument`) which has both `id()` and `mutationToken()` available. both Document and DocumentFragment need to implement this marker interface. Then we can change the signature to `consistent_with(bucket: Bucket, [docs: MutatedDocument...])`.
- The other option would be to just have two overloads: `consistent_with(bucket: Bucket, [docs: DocumentFragment...])` and `consistent_with(bucket: Bucket, [docs: Document...])`. To make the fallback logic work, I think we'd still need to have the id() exposed on the fragment.

@brett19  and I discussed also leaving out the id() on the marker interface, but then the code needs to jump through some hoops (detect at runtime its a Document) to work with the fallback logic. I think just having both on the marker IF is not too bad. And having the id on the fragment is probably helpful in some scenarios.

So I'm up for option 1, and if a language doesn't support marker interfaces (or somethign similar) the fallback would be two overloads? is there any language affected of that? @couchbaselabs/sdk-team wdyt?

mnunberg:
`DocumentFragment` is declared to be a subclass of `Document`. Note that while I didn't define `DocumentFragment` as a concrete object, it is defined as being "like a document".

If we actually need to go and codify rules, the subdoc RFC might need to be modified accordingly.just

Also I think we're forgetting an option 3 which is to accept whatever the actual base class for `Document` and `DocumentFragment` is. A new internal type, `DocumentHeader` would work just fine.

simonbasle:
@mnunberg I don't think this is a great idea making `DocumentFragment` extend `Document`, because that means users could try to pass in a `DocumentFragment` in top-level methods (in the Java SDK all of them accept a `Document` as input)...

brett19:
@simonbasle Indeed.  When I updated the RFC, I believe I removed all instances of the RFC mentioning that DocumentFragment inherited from Document.  The inheritance here would indicate some things which are simply not true.  We should however have a base class or interface to allow you to blindly pass both DocumentFragment's and Document's, as well as aggregate them.

daschl:
Alright, so we need to move forward on that. Any objections @simonbasle @brett19 @mnunberg @brett19 @jeffrymorris @avsej on me updating this spec with the Marker interface? Also, objects on the `MutatedDocument` if name which contains both id and content?

simonbasle:
yep, let's add the marker interface and move that forward
:+1:

jeffrymorris:
I am ok with either option. For the record, a marker interface has no methods, properties or fields defined it's just flag that an instance of a Type should be treated "specially"...that anal-ness aside, either (overload or interface) should work. TBH it's probably best to leave the decision of the interface vs overload up to the implementer since from a user perspective the behavior/usage is the same.

+1

daschl:
While implementing it, the interface felt really weird. I went back to the original proposal which now has 2 overloads, one for document and the other for the subdoc fragment. That should be also easier for the user since they don't need to grasp another type on the method signature.

I also updated the server side API since it changed a bit (from obj to array on the non-1024 case).

mnunberg:
For Python, the form would be something like:

`cb.consistent_with(cb, *rvs)` where `rv` is any `Result` object or subclass

For C, the API remains:
```c
LIBCOUCHBASE_API
lcb_error_t
lcb_n1p_setconsistent_token(lcb_N1QLPARAMS *params,
    const char *keyspace, const lcb_MUTATION_TOKEN *st);

LIBCOUCHBASE_API
lcb_error_t
lcb_n1p_setconsistent_handle(lcb_N1QLPARAMS *params, lcb_t instance);
```

brett19:
Updated RFC: https://gist.github.com/brett19/576a71d9953a1dd626a6

simonbasle:
(on a side note, creating the drafts in a separate branch in **the couchbaselabs original** repo, as opposed to doing so in a fork, has the added benefit that if the forker is not available :sunny: :cocktail:, anybody in the team can still push to the draft branch and move it forward without having to do a gist/another fork & PR :stuck_out_tongue_winking_eye: )

mnunberg:
fwiw, Python:

```python
class MutationState(object):
    def __init__(self):
        self._sv = {}

    def _add_scanvec(self, mutinfo):
        vb, uuid, seq, bktname = mutinfo
        self._sv.setdefault(bktname, {})[vb] = {
            'value': seq,
            'guard': str(uuid)
        }

    def encode(self):
        return couchbase._to_json(self._sv)

    @classmethod
    def decode(cls, s):
        d = couchbase._from_json(s)
        o = MutationState()
        o._sv = d
        # TODO: Validate

    def add_result(self, rv, quiet=False):
        mi = rv._mutinfo()
        if not mi:
            if quiet:
                return False
            raise MissingTokenError.pyexc(
                message='Result does not contain token')
        self._add_scanvec(mi)
        return True

    def add_all(self, bucket, quiet=False):
        added = False
        for mt in bucket._mutinfo():
            added = True
            self._add_scanvec(mt)
        if not added and not quiet:
            raise MissingTokenError('Bucket object contains no tokens!')
        return added

    def __repr__(self):
        return repr(self._sv)

    def __nonzero__(self):
        return bool(self._sv)

    __bool__ = __nonzero__

class N1QLQuery(object):
    def consistent_with(self, state):
        if self.consistency not in (CONSISTENCY_NONE, 'at_plus'):
            raise TypeError(
                'consistent_with not valid with other consistency options')

        if not state:
            raise TypeError('Passed empty or invalid state', state)
        self.consistency = 'at_plus'
        self._body['scan_vector'] = state._sv
```

daschl:
Spec updated with @brett19's changes.

So I think only thing is missing are the APIs - or should we also add the language specifics for each `MutationState` version?

I'll go implement that in java right now.

mnunberg:
C has global state tracking for durability purposes (_if_ mutation tokens are enabled). Can we still leave this exposed? I think it would make life easier for our C users, while not making this a mandatory feature.

daschl:
As discussed, this will be merged in and moved into REVIEW state thereafter.

PR #11, opened 2016-01-08
Proposing modifications per java feedback

simonbasle:
@mnunberg @couchbaselabs/sdk-team I've added Java specificities, offered modifications and provided feedback on the subdoc rfc, currently still in draft.

### DIFF
--- a/draft/0002-subdocapi.md
+++ b/draft/0002-subdocapi.md
@@ -55,19 +55,20 @@ requests in the network.
 
 ### Document/Item
 
-This class, used in top-level KV operations, is also employed by the sub-document
-API to return values or indicate status.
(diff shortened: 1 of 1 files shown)


mnunberg:
I don't know if the path and lookup type should be part of the lookup spec. It might be helpful, but for example would be complicated to implement in Python. The context of what's desired would be implicit in the application request anyway; no?

simonbasle:
@mnunberg let's put it as "optional, for convenience" then?

Also I've gone the `MutliLookupResult` and `MultiMutationResult` route for the return types of `lookupIn` and `mutateIn` (MultiMutateResult only bears the cas and mutationToken, no value, failing multimutate would be denoted by a `MultiMutationException` in java).

I'll update this PR asap.

PR #19, opened 2016-01-14
0005_vbucket_retries

brett19:


mnunberg:
I think this is a good proposal; while also noting that this may somewhat break our 'zero latency all the time' feature.

(+1)

brett19:
I don't think it will have any further affect on our 'zero latency all the time' than we already had.  As vbucket takeover has a period of delay associated with it anyways.

ingenthr:
The latency was always advertised as a very small bit on a per-vbucket basis over the course of a rebalance.  The concern is that it's gone up from something measured in single digit milliseconds (that won't affect most app deployments) to several hundred milliseconds (which some app deployments will notice and all demos will notice)

ingenthr:
This moved to [google drive](https://docs.google.com/document/d/1_arBwv6udzIctdaCt6PaW19URm0c0eljsyPPw8I6VkE/edit).

PR #24, opened 2016-01-21
Adding highlighted examples for Java/.NET

simonbasle:
Also highlighted the Python and C examples according to Github Flavored
Markdown syntax.

### DIFF
--- a/rfc/0002-subdocapi.md
+++ b/rfc/0002-subdocapi.md
@@ -325,60 +325,64 @@ In this section, the abstract design parts need to be broken down on the SDK lev
 The C SDK will not use these verbs, and will instead model the subdoc operations
 after the existing `lcb_store3()` API - using `lcb_sdstore3`:
 
-    // Single command
-    lcb_CMDSDSTORE cmd = { 0 };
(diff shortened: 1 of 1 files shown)


simonbasle:
@jeffrymorris can you check the .NET example see if you spot inconsistencies? The subdoc commit in Gerrit is very light and doesn't even show the `IDocumentFragment` nor its implementation, and I'm not sure what you had in mind for the `MutateIn`/`LookupIn` syntax.

I think it would make sense to have the cas and expiry provided through the `IDocumentFragment`, wdyt?

simonbasle:
cc. @mnunberg

mnunberg:
I'm going to merge this now.. if there are any modifications they can be posted in a subsequent PR. In particular I do like the `getIn<T>` - it looks very elegant

mnunberg:
Oh, and you can put names other than `mnunberg` if needed or desired :) - I just put it there as a dummy user!

simonbasle:
@jeffrymorris please still have a look and follow up with a PR on the .NET code if you feel necessary (or send me comments).

PR #26, opened 2016-02-08
Changes to SubDocument API

brett19:


rfc/0002-subdocapi.md
```diff
@@ -112,171 +89,44 @@ support empty paths. All other forms of path verification is done on the server
...
-Functions exactly like `extend_in`. There are some server-side constraints into
-which values can be used for unique operations - currently it is restricted to
-JSON primitives (so no compound items).
+The Subdocument APIs will be exposed through two separate builders which are available through two new top-level Bucket methods `LookupIn` and `MutateIn`.  Additionally, both `MutateIn` and `LookupIn` when executed will result in a `DocumentFragment` object, which represents the top-level results (Status, CAS, MutationToken) and also contains the result of all subdocument operations which were executed.
 
-* Throws `PathExists` if the unique element already exists
-* Throws `CantInsertValue` if the value is not a JSON primitive.
-* Throws `PathMismatch` if the array already contains non-primitive elements
+### LookupInSet
+This builder exposes the creation of a set of lookup operations to be performed.
```
jeffrymorris: In the preceding paragraph, you discuss "two top-level Bucket methods `LookupIn` and `MutateIn`, how do these differ from LookupInSet (or do they)?

rfc/0002-subdocapi.md
```diff
@@ -112,171 +89,44 @@ support empty paths. All other forms of path verification is done on the server
...
 
-> Because there is too much variation in semantics between `add_unique_in` and
-> `extend_in`, they are separate methods rather than a single unified one.
+`LookupInSet.Get(path)`
+`LookupInSet.Exists(path)`
 
-### `counter_in(docid, path, delta, create_parents, cas, ttl, persist_to, replicate_to)`
-Performs an atomic counter operation on a given path.
+### MutateInSet
+This builder exposes the creation of a set of mutation operations to be performed.
```
jeffrymorris: A similar observation to LookupInSet, should this be "MutateIn"? Or are they two different things?

rfc/0002-subdocapi.md
jeffrymorris: Or is like this: MutateInSet Bucket.MutateIn()?

brett19: Like you said.  It's basically a vaguely ambiguous indication of what you said.  The reason for the current ambiguity is there is still some discussion on whether the API should be a builder which is then 'applied' against a bucket, or if it should be a builder returned by a bucket method.

rfc/0002-subdocapi.md
brett19: See above.

daschl:
Overall I like this updated spec, but in some areas it probably needs to be enhanced (after discussion):

 - The toplevel APIs like lookupIn and mutateIn are missing in how they look like (their actual generic API in the spec), like on mutate cas, expiry, durability requirements in addition to the doc id.
 - The PR states `The SDK MUST verify that a path is non-empty for operations which do not support empty paths`  and as a result the list of methods where we need to check need to be included in the spec to be non-ambiguous.
 - I personally find the `Set` name a bit confusing. Why not name it `Builder` if its a builder?
 - Why does the fragment need to expose information on if it was called on mutate or lookup?
 -  I think the `DocumentFragment` needs to be defined closer in the spec (mandatory fields, all methods it should expose).
 - We need to clarify how the user gets the "index" for `item()` since its not clear how to get there from a `path`.
 - On the errors section, I think it needs to be reworked and grouped into top and content level errors.

mnunberg:
I think the changes are good. More specifically, I am _neutral_ on the matter because I have no stake in the builder pattern API.

brett19:
I have updated the PR according to our discussion earlier today.  Note that there is now a hard-deadline for Feb 24th!

rfc/0002-subdocapi.md
```diff
@@ -112,171 +89,61 @@ support empty paths. All other forms of path verification is done on the server
...
+`MutateInBuilder.Remove(path) -> MutateInBuilder`
+`MutateInBuilder.PushFront(path, value, createParents) -> MutateInBuilder`
+`MutateInBuilder.PushBack(path, value, createParents) -> MutateInBuilder`
+`MutateInBuilder.ArrayInsert(path, value) -> MutateInBuilder`
+`MutateInBuilder.AddUnique(path, value, createParents) -> MutateInBuilder`
+`MutateInBuilder.Counter(path, delta, createParents) -> MutateInBuilder`
 
-See "Numeric Errors" below for a list of errors thrown by this operation.
-This method also returns the current value as a `DocumentFragment`
+`MutateInBuilder.Do() -> DocumentFragment`
```
jeffrymorris: It took me a minute understand the purpose of this method was...while the name is not terrible, i think we could do better...i am just not sure what: Execute(...), Send(), Fire(), GetResults(), Dispatch(), Run() etc....Do().

I can live with it, though...

rfc/0002-subdocapi.md
```diff
@@ -354,199 +223,35 @@ lcb_sched_leave(instance);
...
+//value: Frederick
+
+err = res.Content('invalid', &value)
+//err: Sub-document path does not exist
+//value:
 ```
 
 # Unresolved Questions
 
 * Array method naming: `append` and `prepend` vs `extend`?
```
jeffrymorris: I prefer append/prepend myself..."extend" only accurately describes addition imo.

brett19: What about PushFront/PushBack (which is what is currently in the spec).

PR #27, opened 2016-02-11
error fixes and clarify return value of subdoc ops

mnunberg:


simonbasle:
Not sure about the DocumentFragment renaming, but let's keep that as a separate item.

Here are improvements that I'd like to see in the subdoc rfc that came up from implementing current version (v2, first version with builder API) in Java:

## DocumentFragment should have 4 additional methods:
 * `DocumentFragment.Exists(int index) -> Boolean`: same as other Exists, with the index (all results should be queriable both by index or path)
 * `DocumentFragment.Status(string path) -> responseStatus/code`: returns the status code for the operation instead of throwing an exception in languages where Content throws an exception
 * `DocumentFragment.Status(int index) -> responseStatus/code`: same but index-based
 * `DocumentFragment.Size() -> int`: returns the total number of results, which is also the number of spec originally passed in (for convenience).

## `Do()` is a reserved word in some languages
In Java notably. So I made it `doMutate()` and `doLookup()` for now in the respective builders.

## MutateInBuilder should have builder methods for CAS, expiry and durability
I think having them on the builder limits the bucket level API quite nicely. The `with` prefix can be used to distinguish specs from document level tunings. All these methods need only be invoked once, but if invoked multiple time in the builder chain, the last invocation is the one taken into account:

 * `MutateInBuilder.WithCas(long cas) -> Builder`: apply the given cas for optimistic locking of the whole set of mutations.
 * `MutateInBuilder.WithExpiry(int expiry) -> Builder`: apply the new expiry to the whole document if non-zero (**question**: it's only taken into account on the server side if the whole mutation set succeeds, correct?).
 * `MutateInBuilder.WithDurability(PersistTo persistTo, ReplicateTo replicateTo) -> Builder`: set durability constraints to observe for the whole set of mutations. Variants with only the persistTo or only the replicateTo can also be offered in languages with overloads.

## Error handling clarification
Document level errors should always be thrown when calling `Do()`, IMHO.
Eg. `DocumentNotJsonException`, `DocumentDoesNotExistException`, `CASMismatchException`.

However, what about execution based (and path-based errors that are only discovered on server side)? Should a `DocumentFragment` always be returned, and the exception triggered by calling  eg. `Content(0)`, or should an exception be thrown from `Do()` in the following cases:

 - for **single lookup** operations? (I'd say throw)
 - for **multi lookup** operations? (I'd say return)
 - for **single mutate** operations? (I'd say throw)
 - for **multi mutate** operations? (the previous spec was pretty clear here, it was throw)

Head first into dev as I was, I made it so single ops (lookup and mutate) always return a fragment.

mnunberg:
For error handling, the way I currently have it in Python is to move _all_ top level errors to exceptions unless it's a `PATH_NOT_FOUND` and the command was a `get_in`.

I think how exceptions are handled would be dependent on the language. Python is exception happy (throwing exceptions for pretty much anything)

rfc/0002-subdocapi.md
```diff
@@ -179,11 +208,12 @@ Be aware that for lookup operations which fail partially in languages which can
 These errors are only received for `counter_in` operations:
 
 * NumberTooBig: Existing number value in document is too big (out of range between `INT64_MIN` and `INT64_MAX`
-* ZeroDelta: Got a zero value for a delta
+* BadDelta: This corresponds to the `PROTOCOL_BINARY_RESPONSE_DELTA_ERANGE`
```
simonbasle: Isn't there also the case where the delta, applied to the current value, makes the counter too big?
Note that in a strongly-typed language like Java, only this case above and the "delta is 0" one can happen. Delta must be a number and cannot be itself too large by way of the type being `long` or equivalent (64bits signed integer).

simonbasle:
@mnunberg so would it make sense to say that:
 - **single `get`**
    * return a `DocumentFragment` with the value in case of `SUCCESS`
    * return a `DocumentFragment` with null content in case of `PATH_NOT_FOUND`
    * throw for any other error
 - **single `exist`** (:warning: this seems different from what you've implemented but would be consistent with get's behavior)
    * return a `DocumentFragment` with content = `true` in case of `SUCCESS`
    * return a `DocumentFragment` with content = `false` in case of `PATH_NOT_FOUND`
    * throw for any other error
 - **single mutations**
    * return a `DocumentFragment` in case of success, single content, most of the time Content(0) is null (except in `counter`)
    * always throw exception in case of failure

And:
 - **multi lookup**
    * always return a `DocumentFragment` in case of success or execution time errors
    * has the `content()` method throw an exception for paths that gave execution time errors
    * throw exceptions only for document-level errors
 - **multi mutation**
    * return a `DocumentFragment` with multiple contents (null for most ops, but could be `counter` values) in case of `SUCCESS`
    * throw a `MultiMutationException` with the index of first problematic mutation in case of execution time error (`SUBDOC_PARTIAL_FAILURE`)
    * throw an exception in case of any other top level error

So the only time where a `DocumentFragment.Content(path)` throws an exception would be in case of multi lookup?

mnunberg:
@simonbasle I've adopted something more similar to Brett's API right now (I'll update the RFC to that effect, actually).. but in summary:

We would need consistent behavior with respect to single and multi operations on the protocol, being forced to emulate one or the other.

**For single ops**:
_iff_ the error is a subdoc error then _emulate_ by throwing a `MultiMutationError`.

**For any op (proto- multi or single)**:

The `SubdocResult` (I refuse to call it "fragment") should _always_ be made available if the result is due to a subdoc (not fulldoc) error. Note that in Python the actual `Result` object itself is already available within the exception object (so if someone really choses too, they could suppress the exception and continue handling the result as normal). I am not sure if this actually makes sense in Java, but it would likely be nice to somehow report this information upon failure.

The issue of whether to throw an exception or not is less about _what to return_ and more about whether the library should have a _disruptive_ behavior when encountering such an error, and as such, your table seems to make perfect sense.

One unfortunate thing about emulating single/multi error codes is that the SDK will not be able to understand (and thus not be able to emulate) future subdoc error codes. Considering there is already [one more error code](http://review.couchbase.org/#/c/60052/) in the works, this has actual relevance.

rfc/0002-subdocapi.md
mnunberg: See http://review.couchbase.org/#/c/60052/ - I plan to fix the current delta handling.

mnunberg:
@couchbaselabs/sdk-team - I've modified this PR to only include error information. I believe we are go here.

PR #29, opened 2016-02-26
subdoc: MultiValue

mnunberg:


brett19:
Why don't we just serialize multiple PushFront / PushBack operations into a single ArrayInsert operation internally?

mnunberg:
The logic for 'serializing' that is non-trivial.... the path would need to be analyzed and the SDK would need to make sure no other operations are affecting the path between various push operations.

jeffrymorris:
I am not sure this (the class MultiValue) needs to be a requirement of the public API, since it can be handled easily with a params list or by passing in a collection or array. What other behavior other than being a data structure does the class implement? Perhaps label as optional?

mnunberg:
It may perhaps be the case that no explicit MultiValue is required. I’ve managed to eliminate it in Python, but I’m not sure if the semantics carry over to other languages as well. I assume both Java and .NET allow specifying variable arguments.

Passing in a collection is ambiguous as it’s uncertain if you want to store the array itself or simply the items in the array.

Perhaps someone else should come up with a different way to specify this. In Python I can express this naturally without a special object.

jeffrymorris:
+1

mnunberg:
https://docs.google.com/document/d/1ZXq-JgWW8ywU03Tx51A3jFTysYmlkv2W4ko0kepb3_M/edit#

PR #30, opened 2016-03-04
0010 - Full Text Search

daschl:


daschl:
@couchbaselabs/sdk-team

rfc/0010-cbft.md
```diff
@@ -0,0 +1,566 @@
...
+| `skip(int skip)` | `from` | JSON Number | - | >  0, if not set no size is used (no default value) |
+
+**Examples:** `.skip(5)`
+
+#### Explain
+The boolean `explain` field triggers inclusion of additional search result score explanations.
+
+| Builder Name | Request Path | Wire Type | Default | Constraints |
+| ------------ | ------------ | --------- | ------- | ----------- |
+| `explain()` | `explain` | JSON Boolean | - | -|
```
avsej: shouldn't it accept some boolean?

simonbasle: chaining in this method when building the query would mean you want `"explain": true` in your request, so the assumed boolean value is `true`.

mnunberg:
1. Per discussion with Marty and Simon, you might want to remove `FuzzyQuery` and replace it with its superset, `TermQuery`
2. The bleve site makes reference to

rfc/0010-cbft.md
```diff
@@ -0,0 +1,1338 @@
...
+
+ - **TermQuery:** A query that looks for **exact** matches of the term in the index (no analyzer, no stemming). Useful to check what the actual content of the index is. Usual better alternative is `MatchQuery`.
+ - **Fuzzy Query:** A fuzzy query is a term query that matches terms within a specified edit distance (Levenshtein distance). Usual better alternative is `MatchQuery`.
+ - **Prefix Query:** The prefix query finds documents containing terms that start with the provided prefix. Usual better alternative is `MatchQuery`.
+ - **PhraseQuery:** A query that looks for **exact** match of several terms (in the exact order) in the index. Usual better alternative is `MatchPhraseQuery`.
+ - **MatchAllQuery:** A query that matches all indexed documents.
+ - **MatchNoneQuery:** A query that matches nothing.
+
+## 2. High Level API Overview
+This section describes how the overall API looks like (to be expanded).
```
jeffrymorris: We seem to be missing the parameter for the index name to be queried. Seems like it should be defined on the SearchQueryRequest since it represents the endpoint and not the parameters.

simonbasle: added that as part of the `Bucket` level API, good catch!

simonbasle:
@mnunberg addressed your 3 points.
I've renamed `minShould()` to `shouldMin()` and indicated this is optional if direct setting of `should.min` is available.

rfc/0010-cbft.md
```diff
@@ -0,0 +1,1355 @@
...
+Here is an example of a String query with some search params, on index `beer-index`:
+
+```java
+bucket.query("beer-index", new StringQuery("description:awesome").boost(2), new SearchParams().limit(5))
+```
+
+If idiomatic by the language, it makes sense to provide static factory methods (or similar helpers) to guide the user in the right direction since there are many different query options available. So for example the `SearchQuery` can contain factory methods in a form like this:
+
+```java
+bucket.query("beer-index", SearchQuery.string("+water +abv:>10"), new SearchParams().limit(5))
```
jeffrymorris: I am not sure we need to explicitly separate SearchParams from the SearchQuery class; you end needing both to create the JSON request body.

simonbasle: it makes sense if you see the `SearchQuery` as the equivalent of the N1QL statement. And when you consider that composing queries (eg. in a `DisjunctionQuery`) has the strong requirement that these subqueries should **not** have any searchParams. So you create the JSON request by taking the `indexName` (for the URL) and both `SearchQuery` and `SearchParams` for the body.

rfc/0010-cbft.md
```diff
@@ -138,38 +136,55 @@ Compound Queries:
...
+The search service is an HTTP endpoint that accepts `POST` requests. The url of the endpoint determine the FTS index to be queried, while the body of the request contains a `JSON` representation of the FTS query to perform:
+
+> **POST** /api/index/`{indexName}`/query
 
 For both bucket and cluster level, the approach is the same. Right now only Bucket level is defined in this RFC.
 
 ```java
-bucket.query(SearchQuery[, SearchParams]);
+bucket.query(String indexName, SearchQuery query[, SearchParams params]);
 ```
```
jeffrymorris: Why not make indexName a property of SearchQuery? It shortens the signature and makes it a tighter API IMO since SearchQuery takes in the responsibility for the index to query.

simonbasle: @jeffrymorris because the index name drives the URL to hit, which means it can only appear once, whereas a `SearchQuery` could be composed with others, eg. in a `BooleanQuery`/`DisjunctionQuery`... What to make of index names in such sub-queries? Putting it at the `Bucket` API level allows for a better separation of concerns IMO

jeffrymorris: @simonbasle can a composite query be composed of sub-queries with different index's? If not, I agree with your statement.

simonbasle: no they can't, the index defines the URL endpoint on which the query should be sent, so by definition there can be only one.

simonbasle:
sorry, I rebased the wrong RFC :'(
compared to latest comment (7 days ago), only the 3 last commits are new (162a535, 7474a36, afe851a)

**TODO** as per our discussion: rename the base class for all search queries from `SearchQuery` to something else, rework the `Bucket` level API to have `SearchQuery` be a "wrapper" like for other queries, containing the index name, the query proper and `SearchParams`

ingenthr:
@jeffrymorris– are you comfortable with that?  seems reasonable to me to follow an SDK idiom if there is one or follow the sdk-rfc otherwise.

ingenthr:
Per 10 May review meeting.

@simonbasle to follow up with Michael on the "more options" question.

@simonbasle to follow up on the question on merging PrefexQuery.  Same thing with `MatchAll` and `MatchNone`

@simonbasle to fix the signoff table :)  @daschl doesn't have a time machine.

@simonbasle to work through the TODOs

@simonbasle to add an examples section where the first example is the most common search example using the travel dataset in Java.  This should include all of the bootstrappy things if anything.  Then everyone else will add their language specifics.

rfc/0010-cbft.md
```diff
@@ -96,7 +96,16 @@ This RFC describes the SDK API and behavior exposure of the Couchbase Server Ful
...
+		- [Set-Up for Examples](#set-up-for-examples)
+		- [Java](#java)
+		- [.NET](#net)
+		- [NodeJS](#nodejs)
+		- [Go](#go)
+		- [C](#c)
+		- [PHP](#php)
+		- [Python](#python)
+		- [Ruby](#ruby)
+- [Resolved Questions](#resolved-questions)
```
ingenthr: Need examples now

- [ ] .NET, need one from @jeffrymorris 
- [ ] Node, need one from @brett19 
- [ ] Go, need one from @brett19 
- [ ] C, need an example from @mnunberg 
- [ ] PHP, need an example from @avsej 
- [ ] Python, need an example from @mnunberg 

We can remove the Ruby one.

ingenthr: Maybe this section should just be called "Questions"  or I would not keep the resolution of the question in the list.

simonbasle: I left it there for posterity but yeah that should be pulled up in the PR (which also acts as a discussion history)

simonbasle:
- [ ] Facet name unresolved question brough up by @brett19 added, to be resolved

I removed the Ruby section and the "resolved questions" section as per matt's comment above

simonbasle:
(for archiving of discussion the resolved questions were:)
> - **I wonder if its possible that the compound queries can be put into one API?**
>
>    > **simon:** `BooleanQuery` is kind of a good middle ground
> - **the response also returns the request, seems a little wasteful on network bandwith?**
>
>    > **simon:** this is actually the request that FTS ended up actually executing. Due to optimizations and parsing, it could be different from the one passed in by the user. But that shouldn't happen with SDKs as they guarantee well-formed requests. Maybe include somewhere as a TRACE log?
> - **Should `MatchAllQuery` and `MatchNoneQuery` be included at all?**
>
>    > **simon:** **YES**, marty indicated that you cannot do specific use cases without them, so they're bound to become useful to someone at some point.

brett19:
@simonbasle Why do we have SearchParams instead of having those methods sit on top of the SearchQuery itself?

simonbasle:
@brett19 I'll rephrase the RFC to get rid of SearchParams (or at least make that an option), since most of the SDKs don't have an equivalent in N1QL either, except Java.

simonbasle:
@brett19 @couchbaselabs/sdk-team I have updated the RFC with @brett19 feedback:

1. the `SearchParams` class has been removed. All top level parameters should now be set on the `SearchQuery` directly, using a fluent API (or similar).
2. the `name` of a facet is not part of the `SearchFacet` anymore, but in the `SearchQuery`. The `addFacets` method has thus been renamed `addFacet`, because this prevents languages like Java to add several facets in one call (vararg). But the fluent API style allows to chain `addFacet` several times in a row, which is more verbose but still acceptable.

daschl:
@couchbaselabs/sdk-team one thought, what about naming the ScanConsistency SearchConsistency? So it doesn't clash with the N1QL one -  since they are not used the same way right now. In SDK 3.0 we can unify fts, n1ql and views under one maybe...

simonbasle:
@daschl I'm ok with either way... I'll let everyone either comment on this suggestion or vote by a reaction on your comment ( 👍 , 👎 or 😕 if no opinion)

PR #35, opened 2016-04-06
Broader REVIEW for RFC3 - Index Management

simonbasle:
This PR serves as the medium for broader review of #6 Index Management.
It will transition it from REVIEW to ACCEPTED status, leaving a window for comments until 2016/4/13 (1 week from now).

### DIFF
--- a/rfc/0003-indexmanagement.md
+++ b/rfc/0003-indexmanagement.md
@@ -4,7 +4,7 @@
  - RFC ID: 0003-indexmanagement
  - Start Date: 2015-12-14
  - Owner: Simon Baslé (@simonbasle)
- - Current Status: REVIEW
+ - Current Status: ACCEPTED
 
 # Summary
 The aim of **RFC3**, **"`Index Management`"**, is to offer a simplified API in the `BucketManager` to list, create and drop indexes, for the majority of use cases. In this sense, it is focused on GSI indexes and assumes a few other givens. For cases that don't fit in this picture, the fallback for the user is to craft appropriate `N1QL` queries (eg. in Java using the `Index` fluent API).
--- a/rfc/0003-indexmanagement.md
+++ b/rfc/0003-indexmanagement.md
@@ -55,19 +55,13 @@ The underlying global design would be to rely on existing `N1QL` verbs and, for 
 parameter and the name of the current Bucket.
 
 Each method will offer a bit of tuning with a couple of parameters, so let's detail a more complete signature for each
-one (syntax is pseudo-code closer to Java).
+one (**syntax is to be considered pseudo-code, close to Java**).
 
 ## Listing Indexes
 Signatures:
 
-### Java
 ```java
 List<IndexInfo> listN1qlIndexes()
-```
-
-### PHP
-```php
-listN1qlIndexes() #-> returns array of associative arrays
 ```
 
 Remarks:
@@ -88,7 +82,7 @@ String state //eg. "online", "pending" or "deferred"/"building"
 String keyspace
 String namespace
 JsonArray indexKey //or equivalent Json representation
-String condition //empty string if the index has no WHERE clause
+String condition // this is the WHERE clause. empty string if the index has no WHERE clause.
 ```
 
 ## Creating Indexes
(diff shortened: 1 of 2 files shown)


simonbasle:
Feedback received during the REVIEW phase:

 - [ ] The `watchIndexes` boolean "`includePrimary`" feels clunky and can always be replaced by the primary name (default "`#primary`") anyway, should probably be removed
 - [ ] With FTS coming, having a differenciating prefix in method names would help (proposed "`n1ql`" as in `createN1qlIndex`, `dropN1qlPrimaryIndex`...)
 - [ ] `IndexInfo` should maybe be optional for languages where value objects are considered more of an hindrance and JSON is a first class citizen (eg. Python, JavaScript...)
 - [ ] The `createN1qlIndex` method should allow for a form of `Expression` to be given in order to specify a `WHERE` clause for the index.

I'll check the items that have been integrated into the RFC via subsequent commit.

simonbasle:
@mnunberg @jeffrymorris I have taken your feedback into account with the above. @jeffrymorris can you please confirm the .NET specifics section, notably, looks good?

simonbasle:
note: commit 7c36bca is the one referred to in previous comment, now rebased on master

simonbasle:
@couchbaselabs/sdk-team it is time for sign off!

Please use the "Add your reaction" feature (in the comment's header toolbar, the smiley face with a + sign) => use the 👍 icon to sign off.

I'll update the document right before squashing and merging.

> PS: if you have implemented the previous state of the RFC (REVIEW) but not the changes that have been made in this PR (ACCEPTED), make sure to open a ticket to track that (and optionally comment with the ticket link here, or in issue #6, or whatever rocks your boat)

ingenthr:
Per meeting on 10 May, @avsej and @mnunberg are going to add the PHP and C implementations to this branch of pull request.

simonbasle:
still need @avsej to provide PHP quick example of how the expected API will look like, @brett19 same for Go.
@couchbaselabs/sdk-team also need your voting on the [comment above](https://github.com/couchbaselabs/sdk-rfcs/pull/35#issuecomment-216960096) for final signoff !!!

ingenthr:
Per 12 May discussion, it looks like we'd like a go sample too.  @brett19 can you add this easily?

- [ ] Need a Go sample from @brett19 
- [ ] @simonbasle going to review the reformatting
- [ ] @avsej need you to fix your signoff, it looks like your commit didn't change the table

rfc/0003-indexmanagement.md
```diff
@@ -63,26 +85,41 @@ String state //eg. "online", "pending" or "deferred"/"building"
 String keyspace
 String namespace
 JsonArray indexKey //or equivalent Json representation
+String condition //empty string if the index has no WHERE clause
```
jeffrymorris: I am not sure what this is? Can we have a better description?

jeffrymorris:
One question regarding the "Condition" field otherwise looks good to me :) and .NET source is updated.

rfc/0003-indexmanagement.md
simonbasle: this is the condition part of the index, when you've built it with a `whereClause` expression. The json returned by querying "system:indexes" can contain a `condition` field. If not present, the SDK object should have the empty string instead.

ingenthr: Per discussion 17 May, we don't need to add to the RFC any further description

ingenthr:
Per 19 May review, it looks like the only thing we need to do have you review formatting @simonbasle

ingenthr:
@simonbasle we noted that the recent examples from @brett19 and @jeffrymorris follow a slightly different format in the review on 19 May.  In particular, .NET and go don't show the where clause.  Can you review and come up with whatever we need to try to achieve parity?

simonbasle:
@ingenthr @jeffrymorris in the .NET implementation, the where clause is assembled from lambda expressions given in place of just the names of the fields to index. I think Jeff found it more idiomatic. @jeffrymorris how much of an effort would it be to support both modes (if it doesn't already)?

@brett19, need your input on Go. Do you foresee a problem in providing an (optional) where clause to the index definition?

rfc/0003-indexmanagement.md
```diff
@@ -174,105 +232,245 @@ For SDKs with configurable logging semantics, a mean of toggling a DEBUG log of
...
+     * an explicit T_GSI or T_VIEW.
+     * When using JSON, specify `"using":"gsi"`
+     */
+    unsigned ixtype;
+} lcb_N1INDEXSPEC;
+
+typedef struct {
+    lcb_N1INDEXSPEC spec;
+    lcb_N1IXMGMTCALLBACK callback;
+} lcb_CMDN1IXMGMT;
```
avsej: @mnunberg 

File https://github.com/couchbase/libcouchbase/blob/2.6.0/include/libcouchbase/ixmgmt.h uses **ix** for index in name, so the structure in this spec, but released headers use just X, like `lcb_CMDN1XMGMT`
https://github.com/couchbase/libcouchbase/blob/2.6.0/include/libcouchbase/ixmgmt.h#L138

I think that the spec have to be corrected at least

PR #44, opened 2016-05-09
at_plus example for PHP (fixes #43)

### LINKED ISSUE #43: Add PHP API for 0004-at_plus RFC
It's one line, so it should be simple to do.

avsej:


rfc/0004-at_plus.md
```diff
@@ -196,6 +196,12 @@ The following user errors need to be covered by the SDKs:
     q.consistent_with(ms)
     ```
  - **PHP:** *not included*
```
daschl: remove the not included, since it is :D

rfc/0004-at_plus.md
```diff
@@ -196,6 +196,12 @@ The following user errors need to be covered by the SDKs:
...
+
+    ```php
+    q.consistentWith($metaDoc)
```
daschl: you sure this is right? It should be a $mutationState, right?

rfc/0004-at_plus.md
```diff
@@ -195,7 +195,14 @@ The following user errors need to be covered by the SDKs:
     ms.add_results(*results)  # A `Result` is the Python equivalent of a Document
     q.consistent_with(ms)
     ```
- - **PHP:** *not included*
+ - **PHP:**
+
+    ```php
+    $ms = new MutationState()
```
daschl: that still doesn't align with the methods specified above, right? No static from methods and "addResults" instead of add"

Not sure why python diverges here, but lets try to be consistent if we can :)

Issue #17, opened 2016-01-13
0010_cbft


brett19:

daschl:
See https://github.com/couchbaselabs/sdk-rfcs/pull/30

daschl:
dun dun duuuuun. accepted!

PR #47, opened 2016-06-30
RFC10 FTS - Review Discussion

simonbasle:
This PR will ultimately move RFC10 (#17) to ACCEPTED state.
The intent for it is to centralize discussion raised during the REVIEW phase, and changes these discussions cause to the rfc.

### DIFF
--- a/README.md
+++ b/README.md
@@ -25,7 +25,7 @@ RFC #  | Description | Owner | Status
 7 | Cluster Level Authentication [\[issue\]](https://github.com/couchbaselabs/sdk-rfcs/issues/13) [\[draft\]](https://docs.google.com/document/d/1CD5OL1ez7euCiLJT91zdWY9R4tW_bkGZ0wsFk1UDtyY/edit) | Brett | REVIEW
 8 | [Datastructures](rfc/0008-datastructures.md) | Mark | ACCEPTED
 9 | 2i Query Support | SDK | IDENTIFIED
-10 | [Full Text Search (FTS) API](rfc/0010-cbft.md) [\[issue\]](https://github.com/couchbaselabs/sdk-rfcs/issues/17) [\[Review PR\]](https://github.com/couchbaselabs/sdk-rfcs/pull/47) | Michael | REVIEW
+10 | [Full Text Search (FTS) API](rfc/0010-cbft.md) | Michael | ACCEPTED
 11 | [Connection String](rfc/0011-connection-string.md) | SDK | ACCEPTED
 12 | ~~Adapt memcached error code handling for future proofing (superseded by RFC 13)~~ | SDK | SUPERSEDED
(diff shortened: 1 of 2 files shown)


simonbasle:
These points were raised by @mschoch

- [ ] I strongly prefer "Query String Query" over "String Query".  I know it sounds redundant, but it is accepted terminology: See https://www.elastic.co/guide/en/elasticsearch/reference/current/query-dsl-query-string-query.html
- [ ] Boolean Field Query says something about "which can reduce index size" which I don't understand.
- [ ] Bleve API does not actually require the style in the highlight request, so "highlight": {} is valid, and uses the default highlighter.  It's not clear if this is just an intentional choice in the SDKs or oversight.
- [ ] Facet size/limit applies to all facets, but is documented in line about Term Facets
- [ ] Document says that all ranges values are inclusive, but this is not correct.  All ranges are inclusive on the lower bound, and exclusive on the upper bound.  Either boundary is optional, but you must specify at least one. Ranges may overlap, though applications should expect documents to be counted in multiple buckets when this is done.

(will be checked as these are resolved by RFC modifications or comments)

- [ ] UPDATE - date query no longer supports changing the date/time parser (need to verify this in 4.5)
  -  [ ] ⚠️ this needs a follow-up
- [ ] field section of wildcard query says it can affect analyzer user, but wildcard query does not use analyzer, same for date range query, numeric range query, regexp query, prefix query
- [ ] search result also can be 412 (precondition failed), this is returned when the requested consistency level could not be satisfied before the timeout was reached

simonbasle:
@couchbaselabs/sdk-team as per Marty's comments, a few changes to the RFC that would warrant changes to the implementations: StringQuery, HTTP 412 error case, highlight style optional.

(the "Reword some pieces per Marty comments" commit doesn't imply implementation changes, unlike what's said in message)

ingenthr:
Each SDK needs to check implementations w.r.t. recent updates:

- [ ] Java
- [ ] .NET
- [ ] Node
- [ ] PHP
- [ ] Python
- [ ] Go
- [ ] C

@couchbaselabs/sdk-team Need everyone to check implementations and check the box please!

brett19:
Hey Simon,
Would it be possible to include the necessary handling details for error 412 in the RFC?
Cheers, Brett

simonbasle:
There should be details in the error section of the document, as visible in
the pull request.

daschl:
@couchbaselabs/sdk-team  heads up! 🔥 a new section on "sort" has been added which is coming in 4.6. 

- [] Java
- [ ] .NET
- [] Node
- [ ] PHP
- [ ] Python
- [] Go
- [ ] C

daschl:
Alright! this looks good, I'll rebase and get this merged as ACCEPTED.

avsej:
For FTS sorting in PHP I opened a [ticket], otherwise PHP is good on this change

[ticket]: https://issues.couchbase.com/browse/PCBC-469

simonbasle:
heyo! glad to see this moving forward :D
it should probably be assigned to someone else though... RFC10, codename "hot potatoe" :rofl:

daschl:
@simonbasle haha thanks didn't realize you were still assigned. I did pick it up and it looks like we can finally drive it to completion ;)

PR #54, opened 2016-09-12
0021 Generic Finds DRAFT -> REVIEW

simonbasle:
This PR hosts the early design discussions during the DRAFT phase of the RFC, ultimately bringing it up to REVIEW status when merged.

### DIFF
--- a/rfc/0021-genericFindQueries.md
+++ b/rfc/0021-genericFindQueries.md
@@ -152,7 +152,7 @@ TBD
 # Changelog
  - 2016-09-07: Initial draft publication
  - 2016-09-08: reworked the matchers/operators idea and builder, removed BETWEEN (added to open questions), added explicit list of "NOT" Operators.
- - 2016-09-15: Re-added between, expressed two alternatives to specify operators (JSON convention and builder API).
+ - 2016-09-15: Re-added between, expressed two alternatives to specify operators (JSON convention and builder API). CONTAINS should be used for the “is value x in the array field y?” case rather than a convenience method over LIKE.
 
 # Signoff
 If signed off, each representative agrees both the API and the behavior will be implemented as specified.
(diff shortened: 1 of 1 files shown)


ingenthr:
@brett19 if you have cycles in the next day or so, it'd be good to consolidate this one a bit.  Note there is also http://review.couchbase.org/#/c/67874/ which has Simon's latest thoughts.

rfc/0021-genericFindQueries.md
```diff
@@ -0,0 +1,162 @@
+# Meta
+
+ - RFC Name: Generic Find Querying
+ - RFC ID: 0021
+ - Start Date: 2016-09-07
+ - Owner: @simonbasle
+ - Current Status: DRAFT
+
+# Summary
+Performing simple selects without writing any N1QL.
```
ingenthr: I'd probably phrase this differently.  It's not so much that avoiding N1QL is a goal but rather to keep the simple things simple.  There are lots of situations where just the predicate is needed.

ingenthr:
POJO mapping, not required.  Might be optional.

ingenthr:
Scenario, an e-commerce application that has user profiles who may be sellers and buyers.

Scenarios: 

1. Look up a user’s document by exact match to their {userid | facebook ID | email address}, return the document.  This will have to consider a type=user.
2. Sub-scenario: Look up the document and unpack it into a user defined class.
3. Look up a set of users who are currently logged in and return a collection
4. Sub-scenario: look up a set of users who are logged in where a sub-object or sub-arrayindicates they’re specialists in selling cranes.  Specialty: {sells: “cranes”}
5. Find the items I have for sale whose sales expire this upcoming weekend.
6. Find the sellers who are in the country FR or AT.
7. Find all sellers from the “couchbase.com” domain.

daschl:
@brett19 @ingenthr how did we end up on this? stalled and closable for now?

PR #55, opened 2016-10-31
Update watch delay increment to 50ms

MikeGoldsmith:


daschl:
@MikeGoldsmith looks like a good improvement, I think we need full +1 from all owners so we can merge this.

mnunberg:
Python:

```
    def n1ql_index_watch(self, indexes,
                         timeout=30, interval=1, watch_primary=False):
```

MikeGoldsmith:
.NET currently uses a fixed 50ms interval
```
private static readonly TimeSpan WatchIndexSleepDuration = TimeSpan.FromMilliseconds(50);
```

daschl:
@couchbaselabs/sdk-team  heads up! 🔥 

- [ ] Java
- [ ] .NET
- [ ] Node
- [ ] PHP
- [ ] Python
- [ ] Go
- [ ] C

daschl:
@MikeGoldsmith the only thing that might make sense is to add an explicit "errata" to the rfc at the end which mentions this change.

mnunberg:
While this might be out-of-scope for this RFC, what's the recommended retry interval for clients which don't use a linearly increasing interval?

MikeGoldsmith:
@daschl @mnunberg I've added a proposed static delay of 200ms if an increasing delay is not possible and added the errata section to describe changes

rfc/0003-indexmanagement.md
```diff
@@ -443,3 +443,8 @@ If signed off, each representative agrees both the API and the behavior will be
 | C        | Mark Nunberg   | 2016/05/12              |
 | Python   | Mark Nunberg   | 2016/05/12              |
 | PHP      | Sergey Avseyev | 2016/05/17              |
+
+# Errata
+1. 18th April 2017
+- Update index watch retry increment from 500ms to 50ms
+- Add static watch index retry interval of 250ms if linear retry increments is not available
```
daschl: 250/200ms?

MikeGoldsmith:
@daschl I believe this is ready for review.

avsej (approved):

dnault (approved):

PR #73, opened 2017-08-30
0024: Fast-Failover SDK

daschl:


daverigby:
Proposal looks reasonable to me. 

One comment is that we have a (undocumented, only used by one or two customers) feature where kv_engine can de-duplicate identical cluster_config maps if it hasn't changed since the last not-my-bucket.

This is described in MB-17506 and was first added in 4.5.0. 

I mention this not for any specific reason, just that in the context of efficiently handling config changes we might want to allow individual clients to opt-in to this feature - currently it's enabled globally for memcached.

Relevant patches: 
* http://review.couchbase.org/58887
* http://review.couchbase.org/60705

daschl:
@daverigby thanks **very much** for the feedback! Actually I think that config deduplication is something we should look across the board in the SDKs since it can help in general. I'll bring it up once all the spock-scrambling is complete :)

PR #79, opened 2017-12-08
Added IPv6 information to Connection String RFC.

brett19:


rfc/0011-connection-string.md
```diff
@@ -47,6 +47,9 @@ _couchbase._tcp.* (for non-secured memcached)
 _couchbases._tcp.* (for SSL-secured memcached)
 
 
+### IPv6 Addresses
+IPv6 addresses will be specified following the IETF RFC-2732 document.  The literal address will be enclosed within "[" and "]" characters to disambiguate a IPv6 octal separator and an explicit port.  Note that IPv4 addresses written in IPv6 notation (i.e. ::192.9.5.5, ::ffff:192.0.2.128) should still be resolved as IPv4 addresses within a client.
```
ingenthr: Technically speaking RFC-2732 is about URLs and RFC-5952 is about textual IP address representation.  Since connstr doesn't conform to URL in all cases (the comma one), better add RFC-5952 to the list.

rfc/0011-connection-string.md
```diff
@@ -64,9 +67,12 @@ All key/value pairs should use lower-case, underscore seperated names.
   - http://fqdn
   - http://fqdn?key=value
   - couchbases://fqdn
+  - couchbase://[3ffe:2a00:100:7031::1]
```
ingenthr: Should add an IPv6 example after line 64 and line 65 above, right?  These are the RFC-5952 examples.

rfc/0011-connection-string.md
brett19: I think RFC-5952 is a given here, there is no other method for specifying an IPv6 address that any reasonable (or unreasonable really) person might try to use (note that we don't refer to the IPv4 RFC here either :D).
I specified RFC-2732 as it is relatively generic and only specifies how to encode an IPv6 address for use as a hostname, nothing which conflicts between the URL and Connection Spec RFCs.  I additionally included the most relevant portion of RFC-2732 beneath it to reduce clicks-to-understanding.

rfc/0011-connection-string.md
brett19: What additional examples would I be adding?

ingenthr: There are no examples of how you'd do comma separated multiple hosts with IPv6, but there are examples with dotted quads and even dotted quads with port numbers.

brett19:
Updated to correctly indicate that the IPv6 implementation is similar but not identical to the IETF RFC.  Included some additional new examples.

brett19:
I didn't want to reference RFC-5952 as it only mentions the "[]" version in an arbitrary example, but then references RFC-3986 for the specifics on how that logic works (which is what RFC-2732 is actually updating).

daschl:
@brett19 good to go!

PR #95, opened 2019-11-01
Resolve inconsistencies in description of signature calculation

dnault:
* Specified that the Base64 encoder must emit padding if applicable.
* Removed conflicting/incorrect specification for signature calculation.

### DIFF
--- a/rfc/0032-field-level-encryption.md
+++ b/rfc/0032-field-level-encryption.md
@@ -283,10 +283,11 @@ Where:
 | ------------- |:-------------:|
 |PREFIX | Prefix for temporary field that holds content of encrypted field. This should be configurable so that naming conflicts can be easily resolved at the application level. The default is “crypt”.|
 |FIELDNAME |The original name of the field that contained the data to be encrypted |
+|BASE64() | Standard Base64 encoding. The results MUST include padding if applicable. Padding is required because the encoded results are included in signature calculation, which must be consistent across implementations.|
 |“kid”| The “key-identifier” for resolving the key used to decrypt/encrypt from the KeyStore. See section on Key Management above.|
 |“alg” |The algorithm used to encrypt/decrypt. |
 |“ciphertext” | A Base64 encoded string that is the value from the field that has been encrypted.|
-|“sig”|Optional, required for AES. The HMAC signature of the following fields concatenating and then base64 encoding them: “alg”, “iv”, and “ciphertext”. Order is important and must be respected across all implementations |
+|“sig”|Optional, required for AES. The HMAC signature, calculated as described above. |
 |“iv”|Optional, required by AES. The Base64 encoded initialization vector used for the cipher-text. |
 
 The purpose for the temporary field for storing the encrypted data is so that we store the encrypted data as a string; if we chose to store in the original field, there may be a type-mismatch if the field type was say a integer. For strongly typed languages this may cause an exception or error when deserializing and the string value is assigned to its original property.
@@ -728,16 +729,24 @@ There are a few items out of scope at this time for v1.0, however they very much
   * Algorithm upgrading - for example going from AES-256 to AES-512
   * Support for KMIP based key stores
 
+
+## Changelog
+
+YYYY-MM-DD Initial draft publication
+
+2019-10-31
+* Specified that the Base64 encoder must emit padding if applicable.
+* Removed conflicting/incorrect specification for signature calculation.
+
+
 ## Sign Off
 
 |Language |Representative |Date |
-|:--- |:--- |:--- |--- |
+|--- |--- |--- |
 |C|Sergey Avseyev|2018-06-25|
 |Go|Brett Lawson|2018-06-25|
 |Java|Subhashni Balakrishnan| - |
 |.NET|Jeff Morris|2018-06-22|
 |NodeJS|Brett Lawson|2018-06-25|
 |PHP|Sergey Avseyev|2018-06-25|
-|Python|Ellis Breen|22/6/2018|
-
-
+|Python|Ellis Breen|2018-06-22|


rfc/0032-field-level-encryption.md
```diff
@@ -283,10 +283,11 @@ Where:
 | ------------- |:-------------:|
 |PREFIX | Prefix for temporary field that holds content of encrypted field. This should be configurable so that naming conflicts can be easily resolved at the application level. The default is “crypt”.|
 |FIELDNAME |The original name of the field that contained the data to be encrypted |
+|BASE64() | Standard Base64 encoding. The results MUST include padding if applicable. Padding is required because the encoded results are included in signature calculation, which must be consistent across implementations.|
 |“kid”| The “key-identifier” for resolving the key used to decrypt/encrypt from the KeyStore. See section on Key Management above.|
 |“alg” |The algorithm used to encrypt/decrypt. |
 |“ciphertext” | A Base64 encoded string that is the value from the field that has been encrypted.|
-|“sig”|Optional, required for AES. The HMAC signature of the following fields concatenating and then base64 encoding them: “alg”, “iv”, and “ciphertext”. Order is important and must be respected across all implementations |
+|“sig”|Optional, required for AES. The HMAC signature, calculated as described above. |
```
jeffrymorris: This would be a breaking change I believe between SDK2 and SDK3, would it not?

rfc/0032-field-level-encryption.md
```diff
@@ -283,10 +283,11 @@ Where:
 | ------------- |:-------------:|
 |PREFIX | Prefix for temporary field that holds content of encrypted field. This should be configurable so that naming conflicts can be easily resolved at the application level. The default is “crypt”.|
 |FIELDNAME |The original name of the field that contained the data to be encrypted |
+|BASE64() | Standard Base64 encoding. The results MUST include padding if applicable. Padding is required because the encoded results are included in signature calculation, which must be consistent across implementations.|
```
jeffrymorris: I think we elected to not support padding sometime back?

rfc/0032-field-level-encryption.md
brett19: The 'as described above' is reflecting that there is a description of this field directly above the table already.  We should make sure that the algorithm reflected above actually mirrors the implementation that ended up in the RFCs though.

rfc/0032-field-level-encryption.md
brett19: I agree that we should clarify whether the base64 encoding should be padded or not in this RFC, but we do need to identify which strategy we used in the original implementation and use that.  @dnault can you confirm that the 'MUST include padding' text reflects the existing implementations?

rfc/0032-field-level-encryption.md
```diff
@@ -728,16 +729,24 @@ There are a few items out of scope at this time for v1.0, however they very much
...
+2019-10-31
+* Specified that the Base64 encoder must emit padding if applicable.
+* Removed conflicting/incorrect specification for signature calculation.
+
+
 ## Sign Off
 
 |Language |Representative |Date |
-|:--- |:--- |:--- |--- |
+|--- |--- |--- |
```
brett19: Why are you removing the alignment specifiers?

rfc/0032-field-level-encryption.md
```diff
@@ -728,16 +729,24 @@ There are a few items out of scope at this time for v1.0, however they very much
   * Algorithm upgrading - for example going from AES-256 to AES-512
   * Support for KMIP based key stores
 
+
+## Changelog
+
+YYYY-MM-DD Initial draft publication
```
brett19: YYYY-MM-DD should be specified and can be determined vaguely based on when this was first contributed to Git.

rfc/0032-field-level-encryption.md
brett19: For reference, the Node.js implementation is:
BASE64(HMAC_SHA256(hmacKey, kid + alg + iv + ciphertext))
Hopefully this is reflected by other SDKs considering it wouldn't make a lot of sense to be BASE64 encoding the contents of an HMAC signature, they should just be concatenated.

rfc/0032-field-level-encryption.md
brett19: For reference, the Node.js implementation does pad the base64, however based on the above-mentioned implementation of the signature generation, and the fact that the signature is used as a top-level construct, it shouldn't actually matter to the SDK if it is padded or not.  We should probably specify a "MUST include padding" or "MUST exclude padding" on the encode side, and a "MUST not care about padding" on the decode side.

jeffrymorris: We do use padding.

rfc/0032-field-level-encryption.md
dnault: I agree it makes no sense to Base64 encode the signature components. Unfortunately, the Java implementation Base64 encodes the IV and ciphertext when calculating the signature. It's probably doing this because the RFC says:

>The value of sig can be computed as follows:
>
>```
>sig = BASE64(HMAC256(kid + alg + BASE64(iv) + BASE64(ciphertext))))
>```

For what it's worth, the example signatures in [this section of the RFC](https://github.com/couchbaselabs/sdk-rfcs/blob/master/rfc/0032-field-level-encryption.md#ciphers) can only be matched if the IV and ciphertext are Base64 encoded when calculating the signature.

dnault: I think the way forward is probably to update the RFC to match what .Net and Node.js are doing. The Java implementation can be updated to generate the correct signature but accept both versions.

rfc/0032-field-level-encryption.md
dnault: Because I am dumb and didn't know that's what they were. The important change on this line is to remove the extra  column, since the table data only has 3 columns, and putting 4 columns in the header caused the table to render incorrectly.

rfc/0032-field-level-encryption.md
jeffrymorris: @dnault - that is a bug in the RFC, it should be `sig = HMAC256(BASE64(kid + alg + iv + ciphertext))`

dnault:
Closing as "invalid" since the edit I proposed is not consistent with how implementations are actually supposed to generate signatures, and @jeffrymorris has volunteered to update the RFC with correct examples.

PR #98, opened 2020-09-22
Various updates to management RFC 

chvck:


avsej (approved):

rfc/0054-sdk3-management-apis.md
```diff
@@ -1090,9 +1090,25 @@ public interface IAnalyticsIndexManager{
...
+     
+     void CreateLink(AnalyticsLink link, CreateLinkAnalyticsOptions options);
+     
+     void ReplaceLink(AnalyticsLink link, ReplaceLinkAnalyticsOptions options);
+     
+     void DropLinkFromScope(string linkName, string scopeName, DropLinkAnalyticsOptions options);
+     
+     void DropLinkFromDataverse(string linkName, string dataverseName, DropLinkAnalyticsOptions options);
+     
+     void GetLinkFromScope(string linkName, string scopeName, GetLinkAnalyticsOptions options);
```
dnault, line 1102: Should these GetLinkFrom* methods return `AnalyticsLink` instead of `void`? And GetAllLinks return `List<AnalyticsLink>`?

chvck: Yes, will update tomorrow.

rfc/0054-sdk3-management-apis.md
```diff
@@ -2523,6 +2863,142 @@ interface AnalyticsIndex{
...
+The object names should be implemented in a way idiomatic to the implementing language.
+
+* `dataverse` (string) - The dataverse that the link belongs to.
+* `scope` (string) - The scope that the link belongs to. Only available in server 7.0+. Form is `bucket/scope`.
+* `name` (string) - The name of the link.
+* `accessKeyID` (string) - The AWS S3 access key.
+* `secretAccessKey` (string) - The AWS S3 secret key.
+* `sessionToken` (string) - The AWS S3 token if temporary credentials are provided. Only available in server 7.0+.
+* `region` (string) - The AWS S3 region.
+* `serviceEndpoint` - The AWS S3 service endpoint.
```
dnault, line 2933: Is `serviceEndpoint` a string?

rfc/0054-sdk3-management-apis.md
```diff
@@ -1494,6 +1510,321 @@ Map where top level keys are dataverse names, and values are a map of dataset na
...
+
+* Required:
+
+None
+
+* Optional:
+
+  * `Timeout` or `timeoutMillis` (`int`/`duration`) - the time allowed for the operation to be terminated. This is controlled by the client.
+  * `Scope` (string) - the name of the scope to restrict links to.
+  * `Dataverse` (string) - the name of the dataverse to restrict links to.
```
dnault, line 1800: Are the `scope` and `dataverse` options mutually exclusive?

rfc/0054-sdk3-management-apis.md
```diff
@@ -2523,6 +2863,142 @@ interface AnalyticsIndex{
...
+The object names should be implemented in a way idiomatic to the implementing language.
+
+* `scope` (string) - The scope that the link belongs to. Form is `bucket/scope`.
+* `name` (string) - The name of the link.
+* `connectionString` (string) - The connection string can be used as an authentication method, connectionString contains other authentication methods embedded inside the string. Only a single authentication method can be used. (e.g. “AccountName=myAccountName;AccountKey=myAccountKey”).
+* `accountName` (string) - The Azure blob storage account name.
+* `accountKey` (string) - The Azure blob storage account key.
+* `sharedAccessSignature` (string) - Token that can be used for authentication.
+* `blobEndpoint` - The Azure blob storage endpoint.
+* `endpointSuffix` - The Azure blob endpoint suffix.
```
dnault, line 2961: what type are `blobEndpoint` and `endpointSuffix`?

rfc/0054-sdk3-management-apis.md
```diff
@@ -2523,6 +2863,142 @@ interface AnalyticsIndex{
...
+The `LinkType()` implementation must return a value that corresponds to `"s3"`.
+
+### AzureBlobExternalAnalyticsLink
+
+`AzureBlobExternalAnalyticsLink` provides a means of mapping external azure analytics link details into an object.
+The `AzureBlobExternalAnalyticsLink` is available in server 7.0 Developer Preview mode and must be marked as `volatile` API stability.
+The following fields are named according to how they are sent when form/json encoded.
+The object names should be implemented in a way idiomatic to the implementing language.
+
+* `scope` (string) - The scope that the link belongs to. Form is `bucket/scope`.
```
dnault, line 2954: Does `AzureBlobExternalAnalyticsLink` have a `dataverse` property? If not, do we understand why?

chvck: See the paragraph just above the fields for this link type. Azure support is 7.0 DP and dataverse is deprecated in 7.0, therefore I don't think that we should support dataverse for this link type. Note that this link type is also marked volatile.

rfc/0054-sdk3-management-apis.md
chvck: Yes, added a note to specify this.

rfc/0054-sdk3-management-apis.md
```diff
@@ -1090,9 +1090,19 @@ public interface IAnalyticsIndexManager{
...
+
+     void ReplaceLink(AnalyticsLink link, ReplaceLinkAnalyticsOptions options);
+
+     void DropLink(string linkName, string dataverseName, DropLinkAnalyticsOptions options);
+
+     List<AnalyticsLink> GetAllLinks(GetAllLinkAnalyticsOptions options);
 }
 ```
 
+NOTE: Languages which support constructs like `Either<>` can implement single functions for `DropLink` and `GetLink`, as is idiotmatic.
```
programmatix, line 1104: GetLink has gone now

rfc/0054-sdk3-management-apis.md
```diff
@@ -2523,6 +2718,141 @@ interface AnalyticsIndex{
...
+* `hostname` (string) - The hostname of the target couchbase cluster. This is `activeHostname` in the returned JSON payload for a get.
+* `encryption` (CouchbaseAnalyticsEncryptionSettings) - The encryption settings for the link. The contents of this field are flattened to the top level of the payload.
+* `username` (string) - The username to use for authentication.
+* `password` (string) - The password to use for authentication.
+
+A value for `type` of `couchbase` must be sent in the form payload, this is returned in the same manner in the JSON payload.
+This field is not exposed on the object and is either inferred from the object type or by calling `LinkType()`.
+
+The `Validate()` implementation must verify that:
+* One, and only one, of `dataverse` or `scope` is set.
```
programmatix, line 2754: No longer relevant I think?

rfc/0054-sdk3-management-apis.md
```diff
@@ -2523,6 +2718,141 @@ interface AnalyticsIndex{
...
+  * Either both `username` and `password` are set OR both `clientCertificate` and `clientKey` are set.
+
+The `LinkType()` implementation must return a value that corresponds to `"couchbase"`.
+
+### CouchbaseAnalyticsEncryptionSettings
+
+`CouchbaseAnalyticsEncryptionSettings` are the settings available for setting encryption level on an analytics link.
+
+* `encryption` (AnalyticsEncryptionLevel) - The encryption level to apply.
+* `certificate` (List<Byte>) - The certificate to use for the encryption when encryption level is set to "full".
```
programmatix, line 2771: Maybe worth calling serverCertificate to differentiate it?

rfc/0054-sdk3-management-apis.md
```diff
@@ -1494,6 +1504,180 @@ Map where top level keys are dataverse names, and values are a map of dataset na
...
+    PUT http://localhost:8095/analytics/link/<scope>/<linkname>
+
+## DropLink
+
+Drops an existing link from a scope.
+
+### Signature
+
+```
+void DropLink(string linkName, string scopeName, [options])
```
programmatix, line 1593: Definitely want it this way round rather than scopeName then linkName?

rfc/0054-sdk3-management-apis.md
```diff
@@ -1494,6 +1506,180 @@ Map where top level keys are dataverse names, and values are a map of dataset na
 
     GET http://localhost:8095/analytics/node/agg/stats/remaining
 
+## CreateLink
```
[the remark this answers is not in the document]
programmatix: Fair enough

rfc/0054-sdk3-management-apis.md
```diff
@@ -1494,6 +1504,180 @@ Map where top level keys are dataverse names, and values are a map of dataset na
...
+
+* Required:
+
+None
+
+* Optional:
+
+  * `Timeout` or `timeoutMillis` (`int`/`duration`) - the time allowed for the operation to be terminated. This is controlled by the client.
+  * `Scope` (string) - the name of the scope to restrict links to.
+  * `Name` (string) - the name of the link to fetch.
```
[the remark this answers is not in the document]
chvck: That's why I added GetLink initially and removed name from these options. This is how the REST API works though... We could rename GetAllLinks to GetLinks?

rfc/0054-sdk3-management-apis.md
```diff
@@ -1494,6 +1504,180 @@ Map where top level keys are dataverse names, and values are a map of dataset na
...
+
+Nothing
+
+### Throws
+
+* `LinkAlreadyExistsException`
+
+* `AnalyticsScopeNotFoundException`
+
+* `InvalidArgumentsException`
```
programmatix, line 1537: Also seen:
Unknown analytics error: 24104:Cannot connect to host for link `travel-sample`.inventory.remoteInventory: java.net.ConnectException: Connection refused (Connection refused)

rfc/0054-sdk3-management-apis.md
```diff
@@ -2523,6 +2754,158 @@ interface AnalyticsIndex{
...
+  * `certificate` is set.
+  * Either both `username` and `password` are set OR both `clientCertificate` and `clientKey` are set.
+
+The `LinkType()` implementation must return a value that corresponds to `"couchbase"`.
+
+On fetching a `CouchbaseRemoteAnalyticsLink`:
+  * The `hostname` property is present as `activeHostname`.
+  * The `password` property must be blanked out/left unset.
+  * The `clientKey` property must be blanked out/left unset.
+  * The `dataverse` property must be set by first checking the `dataverse` field in the response body and if empty falling back to the `scope` field.
```
dnault, line 2806: Is it possible for both the `dataverse` and `scope` fields to be present? To put it another way, does it matter which field we look for first?

chvck: I don't believe that it is possible for both to be present so no I don't think that it does matter.

rfc/0054-sdk3-management-apis.md
```diff
@@ -1494,6 +1530,192 @@ Map where top level keys are dataverse names, and values are a map of dataset na
...
+
+  * `Timeout` or `timeoutMillis` (`int`/`duration`) - the time allowed for the operation to be terminated. This is controlled by the client.
+
+### Returns
+
+Nothing
+
+### Throws
+
+* `LinkAlreadyExistsException`
```
dnault, line 1559: Not **Analytics**LinkAlreadyExists / NotFound ?

dnault: Also looks like the existing "aready exists" exceptions don't have "already" in the name... so just `AnalyticsLinkExists`?

dnault: > Not **Analytics**LinkAlreadyExists / NotFound ?

I guess not, because there's already a `LinkNotFound` exception from earlier analytics managment API.

rfc/0054-sdk3-management-apis.md
```diff
@@ -1494,6 +1530,192 @@ Map where top level keys are dataverse names, and values are a map of dataset na
...
+
+* `AnalyticsScopeNotFoundException`
```
dnault, line 1561: There's an existing `DataverseNotFoundException` that seems to be mapped to the same server error code (24034)... Should we reuse it?

https://docs.couchbase.com/server/current/analytics/error-codes.html

rfc/0054-sdk3-management-apis.md
dnault, line 1559: For consistency with the existing annotations, should this be called `LinkExistsException`?

brett19:
Bit strange that these new APIs all hang off the "AnalyticsIndexManager" and are accessed via cluster.analyticsIndexes()... :S

brett19:
There are a bunch of places in the spec where it just says "List" without specifying the type.

brett19:
I'm genuinely curious what "half encryption" even means...

dnault:
> I'm genuinely curious what "half encryption" even means...

"Encryption" is a misnomer. These are actually settings for authentication and transport security.

* NONE: Connect to the remote Couchbase cluster using an unsecured channel.
Send the password in plaintext.
(For development environments where the server uses LDAP auth, I guess?)

* HALF: Connect to the remote Couchbase cluster using an unsecured channel.
Send the password securely using SASL.

* FULL: Connect to the remote Couchbase cluster using a channel secured by TLS.
If a password is used, it is sent over the secure channel.
Requires specifying the certificate to trust.

brett19:
> encryption (AnalyticsEncryptionLevel) - The encryption level to apply.

Should be a field called encryptionLevel.

rfc/0054-sdk3-management-apis.md
```diff
@@ -1494,6 +1530,192 @@ Map where top level keys are dataverse names, and values are a map of dataset na
...
+
+  * `Timeout` or `timeoutMillis` (`int`/`duration`) - the time allowed for the operation to be terminated. This is controlled by the client.
+
+### Returns
+
+Nothing
+
+### Throws
+
+* `AnalyticsScopeNotFoundException`
```
brett19, line 1604: This should be DataverseNotFoundException I think.

rfc/0054-sdk3-management-apis.md
```diff
@@ -1494,6 +1530,192 @@ Map where top level keys are dataverse names, and values are a map of dataset na
...
+The URI is dependent on the options.
+If `dataverse` is not set then:
+
+    GET http://localhost:8095/analytics/link?type=<linktype>
+
+* If `dataverse` is set and contains a `/` then the name is url encoded to escape any `/` within the name and URI is:
+  *  If `dataverse` is set and `name` is not:
+    * `GET http://localhost:8095/analytics/link/<dataverse>?type=<linktype>`
+  *  If `dataverse` is and `name` are both set:
+    * `GET http://localhost:8095/analytics/link/<scope>/<linkname>?type=<linktype>`
```
dnault, line 1712: What is `<scope>`? should it be `<dataverse>`?

rfc/0054-sdk3-management-apis.md
daschl: I wonder if this should be prefixed with AnalyticsLink... since a "link" is a bit ambiguous in the wider context?

rfc/0054-sdk3-management-apis.md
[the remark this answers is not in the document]
programmatix: GetLinks makes sense I think

rfc/0054-sdk3-management-apis.md
```diff
@@ -1190,8 +1204,13 @@ void CreateDataset(string datasetName, string bucketName, [options])
 
   * `DataverseName` (`string`) - The name of the dataverse to use, default to none. If set then will be used as `CREATE DATASET
     dataverseName.datasetName`. If not set then will be `CREATE DATASET datasetName`.
+    * This name may contain one or more `/`, the `/` must be split on and retokenized before sending.
+    * Retokenizing follows the following rule: `bucket.name/scope.name` becomes `` `bucket.name` ``.`` `scope.name` ``.
+    * Note that there can be multiple instance of `/` within a name and all must be retokenized.
 
   * `Timeout` or `timeoutMillis` (`int`/`duration`) - the time allowed for the operation to be terminated. This is controlled by the client.
+  
+  * `LinkName` (`string`) - The name of the link to use, default to none. If set then is applied in the form of `CREATE DATASET datasetName AT linkName`.
```
dnault, line 1213: I think it might be more complicated? The Analytics DDL reference mentions the syntax is different for "remote" links (couchbase) vs "external" links (s3, azure, etc). It looks like for external links we need to say `EXTERNAL DATASET` instead of just `DATASET`. External datasets also have an optional `USING` parameter to specify the path of the dataset relative to the root of the external datastore.

Since a link name along doesn't tell us whether it's a remote couchbase link, i think maybe we need different parameters depending on the link type? Like, `CouchbaseRemoteLinkName` and `ExternalLinkName` & `Using`, where the two link name parameters are mutually exclusive, and `Using` is optional and only allowed if `ExternalLinkName ` is specified.

rfc/0054-sdk3-management-apis.md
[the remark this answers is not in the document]
dnault: @chvck What lead to the removal of `GetLink`? Seems like a very natural way to get a single link by name.

rfc/0054-sdk3-management-apis.md
dnault: Or perhaps we could discover the link type by doing a lookup by name, since we know the link name and dataverse?

chvck: Alternatively to that we could add a `CreateExternalDataset` function? The options for that function would have `Using`.

rfc/0054-sdk3-management-apis.md
```diff
@@ -1090,6 +1090,14 @@ public interface IAnalyticsIndexManager{
      void DisconnectLink(DisconnectLinkAnalyticsOptions options);
 
      map[string]map[string]int GetPendingMutations(GetPendingMutationsAnalyticsOptions options);
+
+     void CreateLink(AnalyticsLink link, CreateLinkAnalyticsOptions options);
+
+     void ReplaceLink(AnalyticsLink link, ReplaceLinkAnalyticsOptions options);
+
+     void DropLink(string linkName, string dataverseName, DropLinkAnalyticsOptions options);
```
programmatix, line 1098: @chvck I think a common requirement will be for code like this work:

analytics.getAllLinks().foreach(link => analytics.dropLink(link))

E.g. at the start or end of testing.
However the API as-is doesn't support this.  dropLink needs a linkName and dataverseName, and the generic AnalyticsLink interface doesn't provide that.  But, they could - at least with the 3 analytics links we know about currently.

rfc/0054-sdk3-management-apis.md
```diff
@@ -1494,6 +1530,190 @@ Map where top level keys are dataverse names, and values are a map of dataset na
...
+None
+
+* Optional:
+
+  * `Timeout` or `timeoutMillis` (`int`/`duration`) - the time allowed for the operation to be terminated. This is controlled by the client.
+  * `Dataverse` (string) - the name of the dataverse to restrict links to.
+  * `Name` (string) - the name of the link to fetch.
+  * `LinkType` (AnalyticsLinkType) - the type of links to restrict returned links to.
+  
+Note: If `Name` is set then `Dataverse` must be set, otherwise an `InvalidArgumentsException` must be thrown.
```
dnault, line 1688: If `name` is set but not `dataverse`, should we fall back to sending `name` as a query parameter? (It seems to work, just wondering if there's a reason not to do it.)

avsej (approved):

PR #107, opened 2021-10-19
Add Rate Limiting error code with causes

daschl:


avsej (dismissed):

rfc/0058-error-handling.md
```diff
@@ -317,6 +317,37 @@ A Note on IDs: The IDs in this RFC are only for organisational purposes and MUST
 
 * Raised when decoding of the data into the user object failed
 
+### 21 RateLimitingFailure
```
programmatix, line 320: Regarding whether to have a single RateLimitingFailure with multiple causes, or separate errors: for me RateLimitedMaxConnections is a bit of an odd man out and should be its own error.  The other three indicate a temporary rate limit that can be resolved by just trying again later - IMO an application would probably handle all 3 the same way, so no need for a separate error there.  For RateLimitedMaxConnections it's more serious and probably going to require some manual resolution, such as running fewer apps or reducing number of KV endpoints.

rfc/0058-error-handling.md
```diff
@@ -317,6 +317,73 @@ A Note on IDs: The IDs in this RFC are only for organisational purposes and MUST
...
+}
+```
+
+Maps to:
+
+* KeyValue
+  * 0x30 RateLimitedNetworkIngress -> NetworkIngressRateLimitReached
+  * 0x31 RateLimitedNetworkEgress -> NetworkEgressRateLimitReached
+  * 0x32 RateLimitedMaxConnections -> MaximumConnectionsReached
+  * 0x33 RateLimitedMaxCommands -> RequestRateLimitReached
```
chvck, line 343: 0x34 is the maximum scope data size status. Doesn't appear to be in the documentation yet.

rfc/0058-error-handling.md
```diff
@@ -317,6 +317,73 @@ A Note on IDs: The IDs in this RFC are only for organisational purposes and MUST
...
+
+Since the failure itself is a very generic error, an additional Set of `RateLimitingReason` should be part of the error that allows the user - if needed - to further inspect/narrow down the cause. Note that it is a set, since it could contain multiple reasons at the same time.
+
+```
+enum RateLimitingReason {
+  NetworkIngressRateLimitReached,
+  NetworkEgressRateLimitReached,
+  MaximumConnectionsReached,
+  RequestRateLimitReached,
+  ConcurrentRequestLimitReached
```
programmatix, line 333: I'm not sure about MaximumConnectionsReached being in RateLimitingFailure.  If it's a pooled service that's maybe accurate, but for KV it's going to be kvEndpoints and/or too many apps provoking this, which will require manual intervention and so would ideally be a QuotaLimitingFailure.
We could return RateLimitingFailure or QuotaLimitingFailure depending on whether it's a dynamic/pooled service or not.  But perhaps that's getting too sophisticated?

daschl: @programmatix as you said it depends on the service and also how the user has set up their infra (you could imagine spinning up app servers on demand in k8s) vs. having them static. I'd vote for what we have is good enough, and if in doubt we can always bring it in with the other exception in the future as well.

avsej (approved):

brett19 (approved):

dnault (approved):

thejcfactor (approved):

jeffrymorris (approved):
LGTM

rfc/0058-error-handling.md
```diff
@@ -317,6 +317,54 @@ A Note on IDs: The IDs in this RFC are only for organisational purposes and MUST
...
+* Note that since there are many different reasons why a request is rate limited, the error context MUST include the reason / specific type of rate limiting cause for debugging purposes.
+
+Maps to:
+
+* KeyValue
+  * 0x30 RateLimitedNetworkIngress
+  * 0x31 RateLimitedNetworkEgress
+  * 0x32 RateLimitedMaxConnections
+  * 0x33 RateLimitedMaxCommands
+* Cluster Manager (body check tbd)
```
jeffrymorris, line 332: Isn't the ClusterManager a SDK2 concept? In SDK3 we broke into different managers for each service:
UserManager Users();
BucketManager Buckets();
QueryIndexManager QueryIndexes();
AnalyticsIndexManager AnalyticsIndexes();
SearchIndexManager SearchIndexes();

https://github.com/couchbaselabs/sdk-rfcs/blob/master/rfc/0054-sdk3-management-apis.md#general-design

daschl: @jeffrymorris the cluster manager is not a API in this case, but a service (it's on the same level as "keyvalue", "query", "search" ... ) .. in this case it means the service running on 8091/18091

PR #112, opened 2022-01-10
Add collections support to query index management

chvck:


daschl (approved):

rfc/0054-sdk3-management-apis.md
```diff
@@ -342,15 +342,26 @@ Iterable<QueryIndex> GetAllIndexes(string bucketName, [options])
...
+  * `CollectionName` - the name of the collection to restrict indexes to.
+  * `ScopeName` - the name of the scope to restrict indexes to.
+    * If either `CollectionName` or `ScopeName` are set then both *must* be set.
 
 ### N1QL
 
+No collections set:
 ```
 SELECT idx.* FROM system:indexes AS idx
 WHERE keyspace_id = "bucketName" AND `using`="gsi"
```
avsej, line 354: It is going to pull collection aware indexes too, right? Because we are matching by bucket name only. Is this expected behavior? Should we document it somehow explicitly?

rfc/0054-sdk3-management-apis.md
```diff
@@ -395,6 +406,22 @@ void CreateIndex(string bucketName, string indexName, []string fields,  [options
...
 
+  * `CollectionName` - the name of the collection to restrict indexes to.
+  * `ScopeName` - the name of the scope to restrict indexes to.
+    * If either `CollectionName` or `ScopeName` are set then both *must* be set.
+
+### N1QL
+
+No collection set:
+```
+"CREATE INDEX `indexName` ON `bucketName`"
```
avsej, line 417: Could we specify if this form indexes all collections or only default one?

rfc/0054-sdk3-management-apis.md
```diff
@@ -2678,9 +2772,29 @@ interface QueryIndex {
     Optional<String> Condition();
     
     Optional<String> Partition()
+    
+    String BucketName();
+    
+    Optional<String> ScopeName();
```
avsej, line 2778: It is kind of inconsistent, because we ask for collection name, but when we retrieve the index back, the user cannot see the collection name back (instead the user should know that it is keyspace). Looks like the implementation details leaking here.

I would propose to define `CollectionName()` method which will return keyspace if the index is scoped, kind of alias of `Keyspace()`, because we are doing same thing here for bucket name, which is keyspace when for collection-less indexes.

avsej (approved):

rfc/0054-sdk3-management-apis.md
chvck: This creates the index only on the default collection.

rfc/0054-sdk3-management-apis.md
chvck: Done

rfc/0054-sdk3-management-apis.md
avsej: I meant could it be recorded explicitly in the RFC? Right now it says

> Fetches all indexes from the server.

rfc/0054-sdk3-management-apis.md
avsej: Right, but it is not clear for the description of the method:

> Creates a new index.

And also the `CollectionName` and `ScopeName` description talk about restricting, but they are not really restricting the index, because by default it is already restricted to default collection.

avsej (approved):

rfc/0054-sdk3-management-apis.md
```diff
@@ -582,6 +666,22 @@ void BuildDeferredIndexes(string bucketName, [options])
...
+### N1QL
+
+No collection set:
+```
+"BUILD INDEX ON `bucketName` (`index1`, `index2`)"
+```
+
+Collection set:
+```
+"BUILD INDEX ON `bucketName`.`scopeName`.`collectionName` (`index1`, `index2`)"
```
avsej, line 682: these two statements are not correct. Look, the method called `BuildDeferredIndexes` and described as "builds all indexes which are currently in deferred state".

Do we need scope/collection here at all?

The correct statement to build all deferred indexes associated with  current bucket will be

```sql
BUILD INDEX ON `{bucket_name}` (
  (
    SELECT RAW name FROM system:indexes 
    WHERE ((keyspace_id = "{bucket_name}" AND bucket_id IS MISSING) 
        OR (bucket_id = "{bucket_name}")) 
        AND state = "deferred"
  )
)
```

And if we still want to restrict this method with scope and collection, then the second statement should be

```sql
BUILD INDEX ON `{bucket_name}`.`{scope_name}`.`{collection_name}` (
  (
    SELECT RAW name FROM system:indexes 
    WHERE bucket_id = "{bucket_name}"
      AND scope_id = "{scope_name}"
      AND keyspace_id = "{collection_name}"
      AND state = "deferred"
  )
)
```

avsej (approved):

avsej (dismissed):

rfc/0054-sdk3-management-apis.md
```diff
@@ -582,6 +666,37 @@ void BuildDeferredIndexes(string bucketName, [options])
...
+
+### N1QL
+
+No collection set, runs build index on only deferred indexes created on the bucket, this does not include indexes created on scopes and collections:
+```
+
+"BUILD INDEX ON `bucketName` (
+  (
+    SELECT RAW name FROM system:indexes 
+    WHERE (keyspace_id = "bucketName" AND bucket_id IS MISSING)
```
daschl, line 681: nit: `( )` are not needed now anymore

thejcfactor: I do not remember if this was mentioned previously, but it looks like this will not work on server versions < 6.5.  I see a syntax error when trying to use a sub-query to serve as the list of indexes to build. :(

`BUILD INDEX ON `default` ((
	SELECT RAW idx.name FROM system:indexes AS idx
	WHERE (
		(`bucket_id` IS MISSING AND `keyspace_id`="default")
		OR `bucket_id`="default"
		) AND state="deferred"
	))
`

rfc/0054-sdk3-management-apis.md
```diff
@@ -2678,9 +2783,30 @@ interface QueryIndex {
     Optional<String> Condition();
```
avsej, line 2783: Keyspace should be removed

avsej:
@chvck, on server 7.1 I see this payload when getting list of indexes:

```json
 {
    "bucket_id": "default",
    "datastore_id": "http://127.0.0.1:8091",
    "id": "e12c8f82c88d1f3",
    "index_key": [
      "`field`"
    ],
    "keyspace_id": "indexcollection_112551505416822",
    "name": "collections_index_112551505388924",
    "namespace_id": "default",
    "scope_id": "indexscope_112551505408772",
    "state": "deferred",
    "using": "gsi"
  }
```

As you can see it is scoped index, but our `GetAllIndexes` does not handle it in this statement:

```
SELECT idx.* FROM system:indexes AS idx
WHERE ((bucket_id IS MISSING AND keyspace_id = "bucketName") OR bucket_id = "bucketName") AND `using`="gsi"
ORDER BY is_primary DESC, name ASC
```

avsej (approved):

rfc/0054-sdk3-management-apis.md
```diff
@@ -342,12 +342,29 @@ Iterable<QueryIndex> GetAllIndexes(string bucketName, [options])
...
+```
+SELECT idx.* FROM system:indexes AS idx
+WHERE ((bucket_id IS MISSING AND keyspace_id = "bucketName") OR bucket_id = "bucketName") AND `using`="gsi"
+ORDER BY is_primary DESC, name ASC
+```
+
+Collection and scope set, returns all indexes for the given collection in the given scope, in the given bucket:
+```
+SELECT idx.* FROM system:indexes AS idx
+WHERE keyspace_id = "collectionName" AND bucket_id= "bucketName" AND scope_id = "scopeName" AND `using`="gsi"
```
dnault, line 360: Should "Collection and scope set" be split into two cases, depending on whether the collection is the default collection? Otherwise, the client never finds indexes in the default collection. Suggestion:

----

Collection and scope set, and CollectionName is "_default". Returns all indexes for the default collection in the given scope, in the given bucket.
```
SELECT idx.* FROM system:indexes AS idx
WHERE ((bucket_id = "bucketName" AND scope_id = "scopeName" AND keyspace_id = "collectionName") OR (bucket_id IS MISSING AND keyspace_id = "bucketName")) AND `using`="gsi"
ORDER BY is_primary DESC, name ASC
``` 

Collection and scope set, and CollectionName is NOT "_default". Returns all indexes for the given collection in the given scope, in the given bucket.
```
SELECT idx.* FROM system:indexes AS idx
WHERE (bucket_id = "bucketName" AND scope_id = "scopeName" AND keyspace_id = "collectionName") AND `using`="gsi"
ORDER BY is_primary DESC, name ASC
```

rfc/0054-sdk3-management-apis.md
```diff
@@ -342,12 +342,29 @@ Iterable<QueryIndex> GetAllIndexes(string bucketName, [options])
...
+ORDER BY is_primary DESC, name ASC
+```
+
+Scope only set, returns all indexes for the given scope, in the given bucket:
 ```
 SELECT idx.* FROM system:indexes AS idx
-WHERE keyspace_id = "bucketName" AND `using`="gsi"
+WHERE bucket_id= "bucketName" AND scope_id = "scopeName" AND `using`="gsi"
```
dnault, line 367: Should this WHERE clause include indexes in the default collection? Suggestion:

```
WHERE ((bucket_id = "bucketName" AND scope_id = "scopeName") OR (bucket_id IS MISSING AND keyspace_id = "bucketName")) AND `using`="gsi"
```

rfc/0054-sdk3-management-apis.md
```diff
@@ -582,6 +666,40 @@ void BuildDeferredIndexes(string bucketName, [options])
...
+  * `CollectionName` - the name of the collection to restrict indexes to.
+  * `ScopeName` - the name of the scope to restrict indexes to.
+    * If either `CollectionName` or `ScopeName` are set then both *must* be set.
+
+### N1QL
+
+This is done by performing 2 querys; one to fetch the deferred indexes to build and the other to build the deferred
+indexes.
+
+No collection set:
```
dnault, line 678: To handle the case where the user explicitly specifies the default collection, should this be:

    No collection set, or collection set to "_default":

Then down below, should the condition for the alternate path be:

    Collection set (and not "_default")

rfc/0054-sdk3-management-apis.md
```diff
@@ -342,12 +342,29 @@ Iterable<QueryIndex> GetAllIndexes(string bucketName, [options])
 * Optional:
 
   * `Timeout` or `timeoutMillis` (`int`/`duration`) - the time allowed for the operation to be terminated. This is controlled by the client.
+  * `CollectionName` - the name of the collection to restrict indexes to.
+  * `ScopeName` - the name of the scope to restrict indexes to.
```
dnault, line 346: NIT: Is it worth mentioning that if `CollectionName` is specified, `ScopeName` must also be specified?

rfc/0054-sdk3-management-apis.md
```diff
@@ -582,6 +666,40 @@ void BuildDeferredIndexes(string bucketName, [options])
...
   * `Timeout` or `timeoutMillis` (`int`/`duration`) - the time allowed for the operation to be terminated. This is controlled by the client.
 
+  * `CollectionName` - the name of the collection to restrict indexes to.
+  * `ScopeName` - the name of the scope to restrict indexes to.
+    * If either `CollectionName` or `ScopeName` are set then both *must* be set.
+
+### N1QL
+
+This is done by performing 2 querys; one to fetch the deferred indexes to build and the other to build the deferred
+indexes.
```
dnault, line 676: Sergey and I chatted and came to the conclusion that the query for the deferred index names could be replaced by a call to getAllIndexes (post-filtered by status = "deferred"), as long as the call to getAllIndexes is limited to the target collection. Using getAllIndexes is less efficient, but potentially simpler to code and maintain. Apart from efficiency, is there another reason to use a special query to find the deferred index names?

dnault (changes requested):
Hi Charlie! Some notes and a question. Thanks.

rfc/0054-sdk3-management-apis.md
chvck: If the user sets the collection name to "_default" then I think that the existing query would work. 
```
((bucket_id = "bucketName" AND scope_id = "scopeName" AND keyspace_id = "collectionName") OR (bucket_id IS MISSING AND keyspace_id = "bucketName"))
```
I think that the latter part of this OR would result in all indexes being returned not only the ones in the _default collection?

dnault: > I think that the latter part of this OR would result in all indexes being returned not only the ones in the _default collection?

Are you referring to `OR (bucket_id IS MISSING AND keyspace_id = "bucketName")` ? This only matches indexes on the default collection, because an index on a non-default collection always has a `bucket_id`.

> If the user sets the collection name to "_default" then I think that the existing query would work.

Please can you try it and confirm? The reason I expect it to fail is that the JSON for an index on the default collection index always looks like this, regardless of the Couchbase Server version:

```json
[
  {
    "datastore_id": "http://127.0.0.1:8091",
    "id": "1fd61fd5f9e1a0d4",
    "index_key": [
      "`foo`",
      "`bar`"
    ],
    "keyspace_id": "my-bucket",
    "name": "mytest",
    "namespace_id": "default",
    "state": "deferred",
    "using": "gsi"
  }
]
```

A query that looks for `bucket_id = "bucketName"` will never see a default index.

dnault (approved):

programmatix:
Can't wait to get these merged so we can immediately deprecate the new fields!

programmatix:
Yea I was being facetious but we really should merge this. It's defacto merged as all SDKs implement it, and my patchset needs to build on it (e.g. I do need to deprecate those fields). I'll+2 it as soon as I get a chance to check the query statements and add them to my own patchset. Likely to be later this week at this point.

brett19 (approved):

PR #116, opened 2022-06-28
CBD-4887: Clarify encoding of index keys on create index

chvck:


rfc/0054-sdk3-management-apis.md
```diff
@@ -380,7 +380,8 @@ void CreateIndex(string bucketName, string indexName, []string fields,  [options
 
   * `indexName`: `string` - the name of the index.
 
-  * `fields`: `[]string` - the fields to create the index over.
+  * `keys`: `[]string` - the keys to create the index over. The SDK must escape each of these keys individually.
+    Note: this means that keywords like ASC/DESC cannot be used.
```
dnault, line 384: Is it worth adding an example that shows user input and the resulting query? (Personally, I'm confused about what it means to escape keys individually, and why this means keywords can't be used).

dnault (changes requested):

avsej (approved):

dnault (approved):

brett19 (approved):

PR #118, opened 2022-09-13
Rfc config profiles

mikereiche:


mikereiche:
I'm just going to merge this so it doesn't get lost.  It's essentially what I had in the Google doc reformatted.

PR #119, opened 2023-02-03
CBD-5091: Adding CollectionQueryIndexManager

programmatix:


chvck (commented):
I appreciate it's more work but I think it's worth including the queries used for each function, especially build deferred. https://github.com/couchbaselabs/sdk-rfcs/pull/112 adds them for the cluster level query index manager. 

If we don't do this it's near guaranteed we'll end up with different implementations.

programmatix:
@chvck yea that's fair.  I omitted them partly because they're identical to the existing queries, just with the keyspace always fully qualified.

programmatix:
Added SQL++ statements.

chvck (approved):
Looks good to me.

dnault (approved):

brett19 (approved):
Approving, but we should ensure that the title of the Commit that gets merged is modified from the PR title (since the PR title has the wrong name).

programmatix:
Changed the PR title which is hopefully now going to be the winner when I do this merge...

@chvck I intentionally left a couple of revision numbers for you to merge your own, earlier changes into, since mine effectively depend on (uh, deprecate) yours.

rfc/0054-sdk3-management-apis.md
```diff
@@ -592,6 +599,325 @@ Nothing
...
+
+CollectionQueryIndexManager appears on the Collection interface, in the form `collection.queryIndexes()`.
+
+All queries will be sent with a `query_context` parameter of "default:`bucket`.`scope`".  
+This is a mandatory parameter for 7.5, and a primary driver for adding this API.
+(It also means that this API cannot be used with servers below 7.0.)
+
+```
+public interface CollectionQueryIndexManager {
+    Iterable<QueryIndex> GetAllIndexes(GetAllQueryIndexOptions options);
```
jeffrymorris, line 619: GetAllQueryIndexIOptions still contains the ScopeName and CollectionName fields, since they were optional at the Cluster level. I can obsolete them and throw exceptions if they are used at this level, but this is a bit awkward.

programmatix: Yea that's correct. It was this or create new options blocks with awkward names. Neither is a perfect solution, and we went with this approach when David raised it in the last meeting.

I've called out here that those parameters need to be deprecated. I haven't called out that an InvalidArgumentException should be raised if they're used with CollectionQueryIndexManager - guess I'd better do another PR.

PR #123, opened 2023-06-14
RFC-0075: Faster Failover and Configuration Push

avsej:


rfc/0075-faster-failover-and-configuration-push.md
```diff
@@ -0,0 +1,150 @@
...
+        +---------------+---------------+---------------+---------------+
+       8| 0x00          | 0x00          | 0x00          | 0x00          |
+        +---------------+---------------+---------------+---------------+
+      12| 0xde          | 0xad          | 0xbe          | 0xef          |
+        +---------------+---------------+---------------+---------------+
+      16| 0x00          | 0x00          | 0x00          | 0x00          |
+        +---------------+---------------+---------------+---------------+
+      20| 0x00          | 0x00          | 0x00          | 0x00          |
+        +---------------+---------------+---------------+---------------+
+      24| 0x42          | 0x00          | 0x00          | 0x00          |
```
avsej, line 62:
```suggestion
      20| 0x00          | 0x00          | 0x00          | 0x00          |
        +---------------+---------------+---------------+---------------+
      24| 0x24          | 0x00          | 0x00          | 0x00          |
```

rfc/0075-faster-failover-and-configuration-push.md
```diff
@@ -0,0 +1,315 @@
...
+
+[https://issues.couchbase.com/browse/MB-57311]: #
+
+### Get Cluster Config with Known Version
+
+[https://review.couchbase.org/c/kv_engine/+/192301]: #
+
+The KV engine introduces a new HELLO flag called `GetClusterConfigWithKnownVersion` with a value of `0x1d`. This flag
+does not change the behavior of the server but allows determining if the node supports epoch-revision fields for the
+`GetClusterConfig` (`0xb5`) operation. If the node acknowledges `GetClusterConfigWithKnownVersion`, then the SDK can use
```
chvck, line 49: If this is done on a per connection basis, rather than gated by a bucket capability to signal the whole cluster supports it, then won't this add quite a lot of complexity to SDKs? Especially in the mixer cluster case.

rfc/0075-faster-failover-and-configuration-push.md
```diff
@@ -0,0 +1,315 @@
...
+
+The previously mentioned `ClustermapChangeNotificationBrief` feature enables the SDK to subscribe all connections for
+configuration updates. These notifications are lightweight and can be deduplicated by the server when the
+`DedupeNotMyVbucketClustermap` option is negotiated.
+
+#### Mixed Clusters
+
+In clusters where there is a mix of nodes with older server versions, meaning that some nodes do not acknowledge
+`ClustermapChangeNotificationBrief`, the respective connection should notify the configuration monitor about its lack of
+support for configuration pushes from the server. As a result, the monitor should utilize the old polling mechanism for
```
chvck, line 223: Does this mean that the SDK will have to continue sending CCCP requests but only to a subset of nodes in a mixed cluster?

rfc/0075-faster-failover-and-configuration-push.md
```diff
@@ -0,0 +1,315 @@
...
+
+### Enhancements in Handling the `NotMyVbucket` Status
+
+Combination of `DedupeNotMyVbucketClustermap` and `ClustermapChangeNotificationBrief` allows to save traffic by not
+sending configuration, if SDK already seen the same revision, and also sends only pair of `Epoch`/`Revision`. So it is
+up to SDK to initiate configuration update once the non-empty payload returned along with `NotMyVbucket` status code.
+
+Several modifications are required in the SDK:
+1. The retry orchestrator should be able to retry an operation based on configuration updates rather than the timer signal.
+2. The configuration monitor should have the ability to throttle configuration requests due to the following reasons:
```
chvck, line 234: What is meant by throttle configuration requests here? Only allow a single configuration request to be in flight at once? If that per node or across the cluster?

rfc/0075-faster-failover-and-configuration-push.md
avsej: I don't think so, because each connection knows its capability at the time the operation is being scheduled.

In fact you can always prepare `GetClusterConfigWithKnownVersion` (i.e. buffer where bodylen is set to `sizeof(uint64_t) * 2`, and fill the version, but when you schedule it to the connection that does not support new feature, you can set body length to zero.

rfc/0075-faster-failover-and-configuration-push.md
avsej: Yes, it will be safe assumption, but it is open for discussion.

rfc/0075-faster-failover-and-configuration-push.md
avsej: yes, one get_config request in-flight, but also it might delay sending the request by adding some extra time to wait before sending request, so that it will absorb more config update requests in big clusters.

rfc/0075-faster-failover-and-configuration-push.md
chvck: I have some concerns over the complexity this is going to add, something will have to monitor how many nodes in the cluster are still using the old mechanism - at least in Go anyway. It seems like it'd be a lot simpler if there was a cluster capability or something that the SDK could use as a signal to switch over.

rfc/0075-faster-failover-and-configuration-push.md
```diff
@@ -0,0 +1,315 @@
...
+Note, that `DedupeNotMyVbucketClustermap` affects `ClustermapChangeNotification` and `ClustermapChangeNotificationBrief`
+features, that described below. In other words, if deduplication enabled, the cluster configuration will be announce for
+the socket connection only once.
+
+### Enforcing Snappy Compression for Cluster Configuration Payloads
+
+[https://review.couchbase.org/c/kv_engine/+/192152]: #
+[https://review.couchbase.org/c/kv_engine/+/192316]: #
+
+The KV engine introduces a new HELLO flag called `SnappyEverywhere` with a value of `0x13`. Once this flag is
```
brett19, line 121: Is there a reason this flag is called 'SnappyEverywhere' and not 'SnappyConfigs'?  It sounds like there is something more to this that we should probably mention as part of the implementation of this flag.

rfc/0075-faster-failover-and-configuration-push.md
```diff
@@ -0,0 +1,315 @@
...
+details about `Duplex` can be found in [kv\_engine/docs/Duplex.md][kv-duplex]. When both flags are negotiated, the
+server will send unsolicited configuration updates to the SDK without expecting any acknowledgement mechanism. While
+this approach proves to have better responsiveness compared to [RFC-0024: Fast Failover][rfc-0024], it also has its own
+drawbacks, such as:
+
+1. The SDK subscribes all connections using HELLO, and during rebalance, all connections will receive all notifications.
+2. In a Lambda scenario, if failover occurs while the SDK process is paused, upon resuming, the SDK must process all
+   updates on all sockets. This process takes unnecessary time, unlike when the SDK polls every 2.5 seconds.
+
+Since version 7.6, the KV engine introduces the HELLO flag `ClustermapChangeNotificationBrief` (`0x1f`). This flag
```
brett19, line 151: Since we have to send all the features at once in the HELLO flag, this implies that we have to request cluster notifications _and_ request the brief version, but on clusters that only support notifications without the brief, we will end up receiving huge payloads of configs.  Unless these flags are mutually exclusive?

rfc/0075-faster-failover-and-configuration-push.md
brett19: I also agree that this seems quite a bit more complicated than necessary.  Perhaps a simpler logic might be that if at least 3 nodes (or all the nodes in the cluster, if less than 3) acknowledge config push, we disable polling, but otherwise use polling.

rfc/0075-faster-failover-and-configuration-push.md
```diff
@@ -0,0 +1,315 @@
...
+1. The retry orchestrator should be able to retry an operation based on configuration updates rather than the timer signal.
+2. The configuration monitor should have the ability to throttle configuration requests due to the following reasons:
+   1. During rebalance, multiple operations may return a `NotMyVbucket` status, triggering a configuration refresh.
+   2. Since `ClustermapChangeNotificationBrief` will cause all connections to subscribe to updates and receive them, it is
+      necessary to account for potential high volumes of updates.
+
+Below is a diagram that illustrates an example of the SDK workflow, where the GET request is waiting for the arrival of
+a new configuration.
+
+```mermaid
```
brett19, line 242: This diagram is a little confusing between step 4 and step 5.  Step 4 doesn't indicate how it is associated with Step 5 (if its associated at all).  IE: Does refreshing the configuration happen in response to the NMV, or is that part of some asynchronous process thats happening.

rfc/0075-faster-failover-and-configuration-push.md
```diff
@@ -0,0 +1,315 @@
...
+2. `DedupeNotMyVbucketClustermap` (`0x1e`). The SDK should be ready that the KV engine will not repeat configuration
+   payload if it already been sent to the socket by any means (`NotMyVbucket` status, `ClustermapChangeNotification`,
+   `GetClusterConfig`).
+
+3. Out-of-Order Execution. `Duplex` (`0x0c`) feature should be always negotiated in HELLO.
+
+4. `ClustermapChangeNotificationBrief` (`0x1f`). The SDK should always subscribe for configuration notifications, if the
+   server supports it, and fallback to polling if it does not.
+
+5. SDK should not emit configuration refresh request if there is one already in-flight. This should be independent of
```
brett19, line 277: There is a race that can occur here.  If we have an in-flight request for a configuration when we receive a new event that requires another configuration, it's possible that the server will reply to the first request with a response that predates the later event.  I think we need to consider the revision information from the event that triggered the refresh request such that if the original configuration request that was in progress returns an older configuration, that we dispatch another config request.

rfc/0075-faster-failover-and-configuration-push.md
```diff
@@ -0,0 +1,315 @@
...
+```
+
+# Language Specifics
+
+## Feature Checklist
+
+1. `GetClusterConfigWithKnownVersion` (`0x1d`). The SDK should always supply current configuration version if the
+   connection has acknowledged feature flag.
+
+2. `DedupeNotMyVbucketClustermap` (`0x1e`). The SDK should be ready that the KV engine will not repeat configuration
```
brett19, line 268: I believe there is a related change to this where NVM's will only contain the revision number rather than the entire configuration.  Should we leverage that to simplify the overall SDK handling of errors?

rfc/0075-faster-failover-and-configuration-push.md
avsej: that still requires counting nodes. Do you mean that we should still use polling and push if there are 5 nodes in cluster, where only 2 nodes support push?

I see it like we still have list of nodes which are used for round-robin, we just need to remove from this list addresses, that ack Push Hello flag.

rfc/0075-faster-failover-and-configuration-push.md
avsej: It uses "Everywhere" because semantics is different. It says "My SDK supports compression Everywhere for ALL packets, not just configuration".

rfc/0075-faster-failover-and-configuration-push.md
avsej: The flags are independent, and I will update paragraph to make it clear that we only send `ClustermapChangeNotificationBrief`. And in general when both flags supported by the KV engine, the brief version "wins".

rfc/0075-faster-failover-and-configuration-push.md
avsej: As it is said below, this flag affects all configuration updates, not only NMV. Could you clarify what do you mean by similification of the error handling?

rfc/0075-faster-failover-and-configuration-push.md
avsej: Right, it makes sense. So I will rephrase this.

rfc/0075-faster-failover-and-configuration-push.md
```diff
@@ -0,0 +1,315 @@
...
+    config_monitor-->config_monitor: wait to throttle config requests
+    config_monitor->>+conn_2: get_config()
+    conn_2->>+kv_node_2: get_config()
+    kv_node_2->>-conn_2: configuration(epoch=1, rev=11)
+    conn_2->>-config_monitor: apply new configuration
+    config_monitor->>retry_orchestrator: purge waiting queue(epoch=1, rev=11)
+    retry_orchestrator->>conn_1: retry get("foo")
+    conn_1->>+kv_node_2: get("foo", vb=115)
+    kv_node_2->>-conn_1: Success()
+```
```
programmatix, line 274: Nb to anyone else wanting to see this visually: https://mermaid.live/edit#pako:eNqNk01vgzAMhv9KlBNo7ADSLkjltOt2qbQTEkqDgaglYYnTClX97zOwIrrSqZwc7Od14o8zl6YEnnIH3x60hHclaivaXDP6hEejfbsDO52l0bqIX7PsZX8sNIFFnLIaMMh5ZUzOI3bcbeL4LZzC5yAiXic2ZZ8GP_qvnZd74qAzstnEEbNwJDC8STPksYC2L4yVDTi0Ao1NWQe6VLoOKHHE5szrUvc8ya6KemSmAzopoxkadiJL14yq4uHuWmRWqi5ao9VIW6gsuIZN__2kMlOLUKL_siehcEiIjTWIB_glSJNyO3SrMtMddJGM9S8mb7CsX7Js02rY7JzbQ2E3L_ivP1fq5i2i6w4903B6ohLZgz7YGsaazNUPnm5tll2nbHQuRzN8NMHJMxOcLCd466UE54KQR7wF2wpV0gKdByLn2EALOU_JLIXd5zzXF4obNmnba8lTtB4i7rtS4HXZeFqJg4PLD_iiNoo

avsej: image

rfc/0075-faster-failover-and-configuration-push.md
```diff
@@ -0,0 +1,337 @@
...
+
+[https://review.couchbase.org/c/kv_engine/+/192152]: #
+[https://review.couchbase.org/c/kv_engine/+/192316]: #
+
+The KV engine introduces a new HELLO flag called `SnappyEverywhere` with a value of `0x13`. Once this flag is
+negotiated, the node will always use the compressed version of the cluster configuration and data type flags will be set
+to `JSON | SNAPPY` (`0x03`).
+
+Note, that the meaning of the flag `SnappyEverywhere` is that SDK expects and properly handles compression for **ANY**
+operation during communication with the KV engine, this is why the flag called "SnappyEverywhere", and "SnappyConfig".
```
dnault, line 131: typo: and "SnappyConfig" -> and **not** "SnappyConfig"

rfc/0075-faster-failover-and-configuration-push.md
```diff
@@ -0,0 +1,337 @@
...
+feature.
+
+The `GetClusterConfig` (`0xb5`) command is explicitly marked as compatible with OoO execution, allowing it to be served
+without waiting for the completion of in-flight operations. Specifically, `GetClusterConfig` will not wait for long
+operations such as mutations with SyncDurability requirements. All current SDKs are expected to be compatible with the
+OoO execution mode, so no changes are expected.
+
+### Cluster Configuration Notification Changes
+
+Prior to server version 7.6, the KV engine had an opt-in feature to push configuration updates to SDKs. This feature
```
dnault, line 148: Does the `ClustermapChangeNotification` feature still exist in 7.6?

avsej: yes, it does. And the RFC has note that it should not be used in HELLO handshake.

rfc/0075-faster-failover-and-configuration-push.md
avsej: Fixed

rfc/0075-faster-failover-and-configuration-push.md
avsej: I've checked and indeed `NotMyVbucket` status either contains full config body (along with revision), or empty body meaning that the config has been seen already.

https://github.com/couchbase/kv_engine/blob/0a84f5b4f2377c9c6553e83091d0ab1b527cfd28/daemon/cookie.cc#L267-L300

I've updated RFC to send request to retry orchestrator in case of empty body

rfc/0075-faster-failover-and-configuration-push.md
avsej: I've updated diagrams

rfc/0075-faster-failover-and-configuration-push.md
avsej: I checked that `NotMyVbucket` actually renders full configuration (if it hasn't been seen by SDK previously), so this paragraph does not apply anymore. The section text is updated.

https://github.com/couchbase/kv_engine/blob/0a84f5b4f2377c9c6553e83091d0ab1b527cfd28/daemon/cookie.cc#L267-L300

rfc/0075-faster-failover-and-configuration-push.md
avsej: Updated mixed mode case

rfc/0075-faster-failover-and-configuration-push.md
```diff
@@ -0,0 +1,364 @@
...
+            conn_1    -->+  kv_node_1:       get("foo", vb=new_vbucket)
+            kv_node_1 ->>-  conn_1:          Success()
+    end
+```
+
+# Language Specifics
+
+## Feature Checklist
+
+1. `GetClusterConfigWithKnownVersion` (`0x1d`). The SDK should always supply current configuration version if the
```
chvck, line 318: In the case that we're using known versions and cccp polling, if the node we've polled returns success but no body should we consider that a successful poll and retry again after the poll period or should we try the next node to see if it has a newer config?

rfc/0075-faster-failover-and-configuration-push.md
```diff
@@ -0,0 +1,364 @@
...
+
+## Library Changes
+
+### Configuration Push
+
+The previously mentioned `ClustermapChangeNotificationBrief` feature enables the SDK to subscribe all connections for
+configuration updates. These notifications are lightweight and can be deduplicated by the server when the
+`DedupeNotMyVbucketClustermap` option is negotiated.
+
+It is not guaranteed that the configuration will be immediately available on the all nodes of the cluster once one of
```
chvck, line 230: What do we do in the case that somehow all nodes report that thy don't have the new revision?

rfc/0075-faster-failover-and-configuration-push.md
```diff
@@ -0,0 +1,364 @@
...
+1. Response handler should tolerate empty response with `NotMyVbucket` status, as the KV engine assumes that the SDK
+   already seen configuration, and there is no newer configuration available. In this case SDK should just retry the
+   operation.
+
+2. If the reponse payload contains body, it contains current configuration, which should be sent to configuration
+   monitor (manager). The SDK should either synchronously apply configuration, create waiting queue for given
+   `epoch`/`revision` pair.
+   Once configuration applied, the SDK must check if new configuration routes the operation to new endpoint or new
+   vbucket on the old endpoint, and *immediately* dispatch operation to new endpoint (or same endpoint in case vbucketID
+   has changed). In any other case, the SDK should send operation to retry orchestrator.
```
programmatix, line 272: I'm not sure I get this part.  Say the SDK is doing 10,000 operations per second per node.  With dedup, just one of those operations will get a NVMB with a config - the other 9,999 will get a NVMB with an empty body.  Why is there a bunch of special case logic for the single operation?  It would be a _lot_ simpler to implement in the JVM clients if we could asynchronously apply the NVMB config body (all config updating is async in JVM world), and send all requests into the RetryOrchestrator but on a tight retry - like a few milliseconds.  That would be pretty easy to implement and those few milliseconds wouldn't seem to impact fast failover at all.

brett19: @programmatix How would you provide determinism that the config had been applies before you attempted to execute the operation again?

programmatix: I haven't pathfound in the JVM clients yet (David's assigned but asked me to look at it while he's away), but IIRC the config mechanism is an async pub-sub.  Something pushes a config and all subscribers will, unless something has gone tragically wrong, get that config in nanoseconds.  So I'm proposing - and I may want to revise this after pathfinding - permitting an alternative setup that's technically non-deterministic, but will be in practice in basically all cases: send the new config to the config publisher, and scheduled the operation to be retried in X milliseconds.  It should be easy to implement, and it's much less risk than introducing a new blocking mechanism to the config publisher that will be challenging to test and will only trigger during rebalances.  And if something odd happens and the config publisher hasn't published by X milliseconds, then all that happens is operations are retried, we get another NVMB, and go around again.  That would be some more traffic to the cluster - but to stress, this is an in-memory pub-sub stream, and it should always have delivered the config within those X millis.

rfc/0075-faster-failover-and-configuration-push.md
```diff
@@ -0,0 +1,364 @@
...
+
+TBD
+
+# Implementation Details
+
+## Protocol Changes
+
+[https://issues.couchbase.com/browse/MB-57311]: #
+
+### Get Cluster Config with Known Version
```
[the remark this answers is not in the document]
chvck: My understanding is that this reduces network traffic - the node responds with no body if it doesn't have a config newer than what we tell it we already have.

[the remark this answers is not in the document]
programmatix: Ah, ok.  My understanding is that we'd be debouncing/throttling config requests on the SDK side anyway.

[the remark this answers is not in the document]
chvck: I think the debouncing logic was removed from the RFC?

rfc/0075-faster-failover-and-configuration-push.md
```diff
@@ -334,6 +334,11 @@ sequenceDiagram
 6. [OPTIONAL] `SnappyEverywhere` (`0x13`). The SDK should be ready that KV engine might send Snappy-compressed payload with any
    of the response types (including push notifications). Check datatype `SNAPPY` (`0x02`).
 
+7. SDK should be able to handle errors from HELLO. It is not critical right now, but in theory it might be happening in
+   future. For example, Sending `ClustermapChangeNotificationBrief` (`0x1f`) without `Duplex` (`0x0c`) will trigger
+   response with the status code `Einval` (`0x04`) and body `{"error":{"context":"ClustermapChangeNotificationBrief needs Duplex"}}`.
+   See [hello\_packet\_executor.cc][kv-engine-hello-error] for more details.
```
programmatix, line 340: Any error from HELLO is surely a terminal one that could only result from an SDK bug, though?  What would our error handling to this be other than to just fast-fail?

rfc/0075-faster-failover-and-configuration-push.md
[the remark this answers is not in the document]
avsej: Yes, this is correct. No debouncing right now, and that known version is an optimization for the reducing payload in case of duplicated requests

rfc/0075-faster-failover-and-configuration-push.md
avsej: Yes, just fast fail. I meant that the SDK should log this case and include this info with the exception if possible.

rfc/0075-faster-failover-and-configuration-push.md
```diff
@@ -0,0 +1,372 @@
...
+      36| 0x08          | 0x07          | 0x06          | 0x05          |
+        +---------------+---------------+---------------+---------------+
+      40| 0x04          | 0x03          | 0x02          | 0x01          |
+        +---------------+---------------+---------------+---------------+
+    GET_CLUSTER_CONFIG command
+    Field        (offset) (value)
+    Magic        (0)    : 0x80 (client request, SDK -> kv_engine)
+    Opcode       (1)    : 0xb5
+    Key length   (2,3)  : 0x0000
+    Extra length (4)    : 0x00
```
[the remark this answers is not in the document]
daverigby: Yes, generally speaking any command which requires "extra" arguments should pass them in the extras section - the value is generally for an actual user document. 

Admittedly GET_CLUSTER_CONFIG _could_ have passed the new args via the value, given it didn't previously have one, but we try to stick to a consistent general approach.

I do note that the KV-Engine docs don't yet cover the new variant of GET_CLUSTER_CONFIG (https://github.com/couchbase/kv_engine/blob/master/docs/BinaryProtocol.md#0xb5-get-cluster-config) - @trondn  could update them please?

rfc/0075-faster-failover-and-configuration-push.md
```diff
@@ -0,0 +1,372 @@
...
+## Feature Checklist
+
+1. `GetClusterConfigWithKnownVersion` (`0x1d`). The SDK should always supply current configuration version if the
+   connection has acknowledged feature flag.
+
+2. `DedupeNotMyVbucketClustermap` (`0x1e`). The SDK should be ready that the KV engine will not repeat configuration
+   payload if it already been sent to the socket by any means (`NotMyVbucket` status, `ClustermapChangeNotification`,
+   `GetClusterConfig`).
+
+3. Out-of-Order Execution. `Duplex` (`0x0c`) feature should be always negotiated in HELLO.
```
dnault, line 326: Would it be better to call this "Server Push" instead of "Out-of-Order Execution"? It's possible to enable Duplex mode without also enabling out-of-order execution.

avsej: You right, here I mean bi-direction communication `Duplex` (`0x0c`), which is not the same as Out-of-Order execution `UnorderedExecution` (`0x0e`). I will update the patch.

rfc/0075-faster-failover-and-configuration-push.md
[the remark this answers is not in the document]
avsej: I've updated diagrams to clarify that epoch/revision should go into extras

rfc/0075-faster-failover-and-configuration-push.md
dnault: I agree with Graham's commentary. "Immediately retry now" isn't feasible for the JVM SDKs.

rfc/0075-faster-failover-and-configuration-push.md
```diff
@@ -0,0 +1,373 @@
...
+    * Completed initial draft.
+
+# Signoff
+
+| Language    | Team Member    | Signoff Date | Revision |
+|-------------|----------------|--------------|----------|
+| .NET        | Jeffry Morris  |              |          |
+| C/C++       | Sergey Avseyev |              |          |
+| Go          | Charles Dixon  |              |          |
+| Java/Kotlin | David Nault    |              |          |
```
dnault, line 361:
```suggestion
| Java/Kotlin | David Nault    |  2023-07-18  |  #1      |
```

dnault (dismissed):

rfc/0075-faster-failover-and-configuration-push.md
```diff
@@ -0,0 +1,373 @@
...
+    Data type    (5)    : 0x00 (RAW)
+    Vbucket      (6,7)  : 0x0000
+    Total body   (8-11) : 0x00000010 (16 bytes)
+    Opaque       (12-15): 0xdeadbeef
+    CAS          (16-23): 0x0000000000000000
+    Epoch        (24-31): 0x0000000000000042 (66 in base-10)
+    Revision     (32-39): 0x0102030405060708 (72623859790382856 in base-10)
+
+If the node has a cluster configuration newer than what is specified in the example, the response will include the new
+configuration in the body with the data type set to `JSON` (`0x01`). Otherwise, the response will have an empty body
```
dnault, line 96:
```suggestion
configuration in the body with the `JSON` (`0x01`) bit set in the data type. Otherwise, the response will have an empty body
```

rfc/0075-faster-failover-and-configuration-push.md
```diff
@@ -0,0 +1,373 @@
...
+        +---------------+---------------+---------------+---------------+
+      12| 0xde          | 0xad          | 0xbe          | 0xef          |
+        +---------------+---------------+---------------+---------------+
+      16| 0x00          | 0x00          | 0x00          | 0x00          |
+        +---------------+---------------+---------------+---------------+
+      20| 0x00          | 0x00          | 0x00          | 0x00          |
+        +---------------+---------------+---------------+---------------+
+      24| 0x42          | 0x00          | 0x00          | 0x00          |
+        +---------------+---------------+---------------+---------------+
+      28| 0x00          | 0x00          | 0x00          | 0x00          |
```
jeffrymorris, line 73: Looking at packet breakdown below, I don't believe these 4 bytes are needed. The breakdown starts at byte 0 and ends at 39, but these empty bytes make the packet length 43 bytes.

rfc/0075-faster-failover-and-configuration-push.md
```diff
@@ -0,0 +1,373 @@
...
+
+      Byte/     0       |       1       |       2       |       3       |
+         /              |               |               |               |
+        |0 1 2 3 4 5 6 7|0 1 2 3 4 5 6 7|0 1 2 3 4 5 6 7|0 1 2 3 4 5 6 7|
+        +---------------+---------------+---------------+---------------+
+       0| 0x80          | 0xb5          | 0x00          | 0x00          |
+        +---------------+---------------+---------------+---------------+
+       4| 0x00          | 0x00          | 0x00          | 0x00          |
+        +---------------+---------------+---------------+---------------+
+       8| 0x00          | 0x00          | 0x00          | 0x00          |
```
jeffrymorris, line 63: Is the byte in position 11 supposed to be 0x10? (Total body (8-11) : 0x00000010 (16 bytes))

rfc/0075-faster-failover-and-configuration-push.md
```diff
@@ -0,0 +1,373 @@
...
+Epoch and revision are signed 64-bit integers encoded in network (big-endian) order and should be encoded as extras.
+
+
+      Byte/     0       |       1       |       2       |       3       |
+         /              |               |               |               |
+        |0 1 2 3 4 5 6 7|0 1 2 3 4 5 6 7|0 1 2 3 4 5 6 7|0 1 2 3 4 5 6 7|
+        +---------------+---------------+---------------+---------------+
+       0| 0x80          | 0xb5          | 0x00          | 0x00          |
+        +---------------+---------------+---------------+---------------+
+       4| 0x00          | 0x00          | 0x00          | 0x00          |
```
jeffrymorris, line 61: Extra length (4) : 0x10 (16 bytes, two int64_t fields in extras)

rfc/0075-faster-failover-and-configuration-push.md
```diff
@@ -0,0 +1,373 @@
...
+    Magic        (0)    : 0x80 (client request, SDK -> kv_engine)
+    Opcode       (1)    : 0xb5
+    Key length   (2,3)  : 0x0000
+    Extra length (4)    : 0x10 (16 bytes, two int64_t fields in extras)
+    Data type    (5)    : 0x00 (RAW)
+    Vbucket      (6,7)  : 0x0000
+    Total body   (8-11) : 0x00000010 (16 bytes)
+    Opaque       (12-15): 0xdeadbeef
+    CAS          (16-23): 0x0000000000000000
+    Epoch        (24-31): 0x0000000000000042 (66 in base-10)
```
jeffrymorris, line 92: 0x4200000000000000

rfc/0075-faster-failover-and-configuration-push.md
```diff
@@ -0,0 +1,373 @@
...
+    Revision     (32-39): 0x0102030405060708 (72623859790382856 in base-10)
```
jeffrymorris, line 93: 0x8070605040302010?

jeffrymorris (changes requested):
A couple of questions regarding the packet layout for GET_CONFIG and CLUSTERMAP_CHANGE_NOTIFICATION. Looks like typos, but is confusing to me.

ingenthr:
Hey @avsej, @brett19, @dnault, et. al.… looks like this is a bit stuck?  What do we need to resolve to get them unstuck?

avsej:
@ingenthr right. I will try to revisit all notes and update the PR

dnault (approved):

PR #127, opened 2023-08-09
Add history retention settings to buckets/collection management

chvck:


rfc/0054-sdk3-management-apis.md
```diff
@@ -3060,15 +3085,15 @@ Nothing
 
 * `InvalidArgumentsException`
 
-* `CollectionAlreadyExistsException`
```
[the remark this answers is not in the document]
chvck: Thanks for catching this. I was very tired yesterday and totally missed that I'd done this, even on double checking it.

rfc/0054-sdk3-management-apis.md
```diff
@@ -3070,6 +3072,46 @@ Nothing
...
+
+### Throws
+
+* `InvalidArgumentsException`
+
+* `CollectionNotFoundException`
+
+* `ScopeNotFoundException`
+
+* Any exceptions raised by the underlying platform
```
programmatix, line 3109: This isn't correct right?  As all exceptions are meant to derive from CouchbaseException in SDK3.

rfc/0054-sdk3-management-apis.md
```diff
@@ -3569,6 +3611,18 @@ enum StorageBackend {
...
+* `HistoryRetentionCollectionDefault` (`bool`) - Whether to enable history retention on collections by default.
+  * URL query field: `historyRetentionCollectionDefault`
+  * If the SDK does not have the ability to differentiate between not set and `false` then a different type should be used instead, suggested:
+    * `enum HistoryRetentionCollectionDefaultSettings { ON, OFF }`
+
+* `HistoryRetentionBytes` (`int`) - The maximum size, in bytes, of the change history that is written to disk for all collections in this bucket.
+  * URL query field: `historyRetentionBytes`
+
+* `HistoryRetentionDuration` (`duration`) - The maximum number of seconds to be covered by the change history that is written to disk for all collections in this bucket.
+  * URL query field: `historyRetentionSeconds` sent as an int seconds value
```
programmatix, line 3623: Do we want to group these 3 fields behind a HistoryRetention object containing bytes and duration?

rfc/0054-sdk3-management-apis.md
```diff
@@ -3070,6 +3072,46 @@ Nothing
...
+
+* `CollectionNotFoundException`
+
+* `ScopeNotFoundException`
+
+* Any exceptions raised by the underlying platform
+
+### URI
+
+    PATCH http://localhost:8091/pools/default/buckets/<bucket>/collections/<scope_name>/<collection_name> -d maxTTL=<maxExpiry>
```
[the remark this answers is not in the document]
programmatix: I'm confused - CollectionSpec now includes history-retention mode, and this section implies that I can update that setting on existing collections - but it's not part of the PATCH?

rfc/0054-sdk3-management-apis.md
```diff
@@ -3599,11 +3653,18 @@ interface CollectionSpec {
...
+    Bool History();
 }
 ```
 
 `MaxExpiry` is the time in seconds for the TTL for new documents in the collection. Omit if not set.
 
+`History` is whether history retention override is enabled on this collection. Omit if not set - not set will default to bucket level setting.
+* On Create/Update this setting should be guarded by the bucket capability of `nonDedupedHistory`
+* If the bucket capability is not supported then a `FeatureNotAvailableException` must be raised.
+  * Suggested supplementary text: "history is unsupported by the server, are you using a bucket with the magma storage engine?"
```
programmatix, line 3666: So no control over bytes & duration at the collection-level?

Is there any way to unset a true/false back to the default state, e.g. default to the bucket level setting?

programmatix, line 3666: The suggested text makes it a little ambiguous whether the feature is, or is explicitly not, supported on Magma.

rfc/0054-sdk3-management-apis.md
chvck: I've just kept consistency with the rest of the RFC there.

chvck:
> FIT testing wise, I think we'll need to use the proxy to check the SDK is sending these fields. Actually checking the history/dedup status of the bucket seems very challenging...

Just get the bucket, the history settings are in the payload.

rfc/0054-sdk3-management-apis.md
```diff
@@ -3070,6 +3072,46 @@ Nothing
...
+void UpdateCollection(CollectionSpec collection, [options])
+```
+
+### Parameters
+
+* Required:
+
+  * `collection`: [`CollectionSpec`](#collectionspec) - specification of the collection.
+
+* Optional:
```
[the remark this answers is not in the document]
chvck: As above, I was just copying what's already in the RFC for other bucket commands.

rfc/0054-sdk3-management-apis.md
```diff
@@ -3070,6 +3072,46 @@ Nothing
...
+
+  * `Timeout` or `timeoutMillis` (`int`/`duration`) - the time allowed for the operation to be terminated. This is controlled by the client.
+
+  * `maxExpiry` (`duration`) - Value for the max expiry of new documents created without an expiry.
```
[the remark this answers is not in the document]
chvck: Unsure what other SDKs have done but Go certainly doesn't.

rfc/0054-sdk3-management-apis.md
chvck: I was avoiding doing that in Go as I've already had to do some hackery for HistoryRetentionCollectionDefault due to false and not set being the same value so adding another object on top would be quite verbose.

rfc/0054-sdk3-management-apis.md
chvck: > So no control over bytes & duration at the collection-level?

Correct.

> Is there any way to unset a true/false back to the default state, e.g. default to the bucket level setting?

I'll check that and update the RFC either way.

chvck: > Is there any way to unset a true/false back to the default state, e.g. default to the bucket level setting?

No, the setting is inherited from the bucket at creation time. Thereafter the setting is owned at the collection level. Omitting `history` at update time will just not alter the setting.

rfc/0054-sdk3-management-apis.md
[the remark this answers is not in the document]
dnault: Absolute expiry does not make sense in the context of collection maxTTL.

For document CRUD operations we can allow either relative or absolute expiry, because the server ultimately converts a relative expiry to to an absolute expiry, and there's a simple and intuitive way to convert a relative expiry to an absolute expiry: just add the relative expiry to the current time, and viola!

For collection maxTTL, you can't say "I want all documents in the collection to expire on January 1st, 2024". That's not how the maxTTL field works. maxTTL is _always relative_ to the time when the document was created/modified.

rfc/0054-sdk3-management-apis.md
```diff
@@ -3068,7 +3070,47 @@ Nothing
 
 ### URI
 
-    POST http://localhost:8091/pools/default/buckets/<bucket>/collections/<scope_name> -d name=<collection_name> -d maxTTL=<maxExpiry>
+    POST http://localhost:8091/pools/default/buckets/<bucket>/collections/<scope_name> -d name=<collection_name> -d maxTTL=<maxExpiry> -d history=<history>
+
+## UpdateCollection
+
+Creates a new collection.
```
dnault, line 3077: Copy & paste error?

rfc/0054-sdk3-management-apis.md
```diff
@@ -3068,7 +3070,47 @@ Nothing
...
+
+* Required:
+
+  * `collection`: [`CollectionSpec`](#collectionspec) - specification of the collection.
+
+* Optional:
+
+  * `Timeout` or `timeoutMillis` (`int`/`duration`) - the time allowed for the operation to be terminated. This is controlled by the client.
+
+  * `maxExpiry` (`duration`) - Value for the max expiry of new documents created without an expiry.
```
dnault, line 3095: > Value for the max expiry of new documents created without an expiry.

<pendantry>
It also affects documents mutated with an expiry, by ensuring the document expiry is no farther in the future than the collection maxTTL.
</pedantry>

rfc/0054-sdk3-management-apis.md
chvck: Indeed!

rfc/0054-sdk3-management-apis.md
```diff
@@ -3599,11 +3653,18 @@ interface CollectionSpec {
...
+    Bool History();
 }
 ```
 
 `MaxExpiry` is the time in seconds for the TTL for new documents in the collection. Omit if not set.
 
+`History` is whether history retention override is enabled on this collection. Omit if not set - not set will default to bucket level setting.
+* On Create/Update this setting should be guarded by the bucket capability of `nonDedupedHistory`
+* If the bucket capability is not supported then a `FeatureNotAvailableException` must be raised.
+  * Suggested supplementary text: "history is unsupported by the server, magma storage engine must be used"
```
programmatix, line 3666: This message seems a bit ambiguous?  Like is it telling me the problem is the server version, or that I'm not using Magma?

rfc/0054-sdk3-management-apis.md
programmatix: Ok, thanks for checking

rfc/0054-sdk3-management-apis.md
chvck: I've had 2 stabs at the wording and come up lacking :), any suggestions?

rfc/0054-sdk3-management-apis.md
```diff
@@ -3831,13 +3892,20 @@ interface ScopeSpec {
...
+  * Adding `UpdateCollection` function.
   
 # Signoff
 
 | Language   | Team Member         | Signoff Date   | Revision |
 |------------|---------------------|----------------|----------|
 | Node.js    | Brett Lawson        | April 16, 2020 | #9       |
-| Go         | Charles Dixon       | April 29, 2021 | #19      |
+| Go         | Charles Dixon       | April 29, 2021 | #25      |
 | Connectors | David Nault         | April 29, 2020 | #9       |
```
dnault, line 3909:
```suggestion
| Connectors | David Nault         | 2023-08-21      | #25      |
```

dnault (approved):

rfc/0054-sdk3-management-apis.md
```diff
@@ -3068,7 +3070,47 @@ Nothing
...
+    POST http://localhost:8091/pools/default/buckets/<bucket>/collections/<scope_name> -d name=<collection_name> -d maxTTL=<maxExpiry> -d history=<history>
+
+## UpdateCollection
+
+Updates an existing collection.
+
+### Signature
+
+```
+void UpdateCollection(CollectionSpec collection, [options])
```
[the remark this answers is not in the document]
chvck: We can add overloads in languages that support them, but in languages that don't I think that we're stuck on this path. We can't even just put History in the options because we need to return it as part of the CollectionSpec for Get.

rfc/0054-sdk3-management-apis.md
```diff
@@ -3599,11 +3653,18 @@ interface CollectionSpec {
     String ScopeName();
 
     Duration MaxExpiry();
+    
+    Bool History();
```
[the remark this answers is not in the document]
chvck: This mirrors the name in the REST API. I'm not against renaming it but I think we tend to try to stay close to the REST interface.

rfc/0054-sdk3-management-apis.md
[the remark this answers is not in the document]
avsej: can we implement UpdateCollection as ideal API now? we could also provide some overrides and deprecate not-so ideal drop/create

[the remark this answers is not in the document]
chvck: > can we implement UpdateCollection as ideal API now? we could also provide some overloads and deprecate not-so ideal drop/create

I'll update the RFC with the new API and mark the old one deprecated. Some SDKs (without overloads) will have to implement the deprecated API for update to be consistent with the other functions.

programmatix (approved):

rfc/0054-sdk3-management-apis.md
thejcfactor: Does `HistoryRentionBytes` and/or `HistoryRentionDuration` need to be set if `HistoryRententionCollectionDefault` is `true`?  Or do we have default values that should be used?

chvck: The docs state that if neither are set then History will be ignored. I've tested and server does not error if HistoryRententionCollectionDefault is true and both bytes and duration are empty. I'd say we leave it to the server.

rfc/0054-sdk3-management-apis.md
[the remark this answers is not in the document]
chvck: Following team meeting last week I have now updated the RFC with the approach that we settled on.

programmatix (approved):
There's still bits I'd prefer be changed (previously discussed), but hey-ho.

rfc/0054-sdk3-management-apis.md
```diff
@@ -3831,19 +3928,27 @@ interface ScopeSpec {
...
 
 | Language   | Team Member         | Signoff Date   | Revision |
 |------------|---------------------|----------------|----------|
 | Node.js    | Brett Lawson        | April 16, 2020 | #9       |
-| Go         | Charles Dixon       | April 29, 2021 | #19      |
-| Connectors | David Nault         | April 29, 2020 | #9       |
+| Go         | Charles Dixon       | 2023-08-21     | #25      |
+| Connectors | David Nault         | 2023-08-21     | #25      |
 | PHP        | Sergey Avseyev      | April 22, 2020 | #9       |
 | Python     | Ellis Breen         | April 29, 2020 | #9       |
```
thejcfactor, line 3948:
```suggestion
| Python     | Jared Casey         | August 31, 2023 | #25       |
```

rfc/0054-sdk3-management-apis.md
```diff
@@ -3831,19 +3928,27 @@ interface ScopeSpec {
...
+  * Adding `History` to `CollectionSpec.
+  * Adding `HistoryRetentionCollectionDefault`, `HistoryRetentionBytes`, `HistoryRetentionSeconds` to `BucketSettings`.
+  * Adding `UpdateCollection` function and update `CollectionManager` interface.
+  * Adding `CreateCollectionSettings` and `UpdateCollectionSettings`.
   
 # Signoff
 
 | Language   | Team Member         | Signoff Date   | Revision |
 |------------|---------------------|----------------|----------|
 | Node.js    | Brett Lawson        | April 16, 2020 | #9       |
```
thejcfactor, line 3944:
```suggestion
| Node.js    | Jared Casey        | August 31, 2023 | #25       |
```

rfc/0054-sdk3-management-apis.md
```diff
@@ -2986,13 +2986,13 @@ It is recommended to try to parse the first form and fallback to the second so t
 
 ```
 public interface ICollectionManager{
-    ScopeSpec GetScope(string scopeName, GetScopeOptions options);
-
     Iterable<ScopeSpec> GetAllScopes(GetAllScopesOptions options);
 
-    void CreateCollection(CollectionSpec collection, CreateCollectionOptions options);
+    void CreateCollection(String scopeName, String collectionName, CreateCollectionSettings settings, CreateCollectionOptions options);
```
DemetrisChr, line 2991: Should the settings here be an attribute of  `CreateCollectionOptions`, considering that they are optional? I could be wrong, but I think in all other operations any optional parameters are part of the options (i.e. the only optional parameter are the options)

programmatix: Personally I argue that when creating a collection you are giving the definition of that collection - though it's fine for that to be the default definition - so it's not an optional parameter.
There's some useful prior art.  Once is that when creating a document you always have to give the content of the document, even though strictly speaking that's optional - valid for a document to be empty/default.
Second is the bucket management API, with createBucket taking a BucketSettings.  (Should have been CreateBucketSettings though, IMO)

DemetrisChr: Ok, that makes sense. I think one difference with bucket management is that there the bucket name is part of the settings, so you can't not provide the bucket settings. Here it is described by the RFC as optional and you are able to simply do `create_collection(scope_name, coll_name)` which just uses the default settings. I guess this is more of an argument for having scopeName and collectionName be part of CreateCollectionSettings though

dnault (dismissed):

rfc/0054-sdk3-management-apis.md
```diff
@@ -3831,19 +3928,27 @@ interface ScopeSpec {
...
+| Connectors | David Nault         | 2023-08-21     | #25      |
 | PHP        | Sergey Avseyev      | April 22, 2020 | #9       |
-| Python     | Ellis Breen         | April 29, 2020 | #9       |
-| Scala      | Graham Pople        | April 30, 2020 | #9       |
+| Python     | Jared Casey         | 2023-08-31 | #25       |
+| Scala      | Graham Pople        | 2023-08-21     | #25      |
 | .NET       | Jeffry Morris       | April 22, 2020 | #9       |
-| Java       | Michael Nitschinger | April 16, 2020 | #9       |
+| Java       | Michael Nitschinger | 2023-08-21     | #25      |
 | C          | Sergey Avseyev      | April 22, 2020 | #9       |
```
avsej, line 3952:
```suggestion
| C          | Sergey Avseyev      | 2023-09-05 | #25       |
```

rfc/0054-sdk3-management-apis.md
```diff
@@ -3831,19 +3928,27 @@ interface ScopeSpec {
...
+| Go         | Charles Dixon       | 2023-08-21     | #25      |
+| Connectors | David Nault         | 2023-08-21     | #25      |
 | PHP        | Sergey Avseyev      | April 22, 2020 | #9       |
-| Python     | Ellis Breen         | April 29, 2020 | #9       |
-| Scala      | Graham Pople        | April 30, 2020 | #9       |
+| Python     | Jared Casey         | 2023-08-31 | #25       |
+| Scala      | Graham Pople        | 2023-08-21     | #25      |
 | .NET       | Jeffry Morris       | April 22, 2020 | #9       |
-| Java       | Michael Nitschinger | April 16, 2020 | #9       |
+| Java       | Michael Nitschinger | 2023-08-21     | #25      |
```
avsej, line 3951: Shall we update Java Owner here?

rfc/0054-sdk3-management-apis.md
```diff
@@ -3831,19 +3928,27 @@ interface ScopeSpec {
...
 | C          | Sergey Avseyev      | April 22, 2020 | #9       |
 | Ruby       | Sergey Avseyev      | April 22, 2020 | #9       |
```
avsej, line 3953:
```suggestion
| Ruby       | Sergey Avseyev      | 2023-09-05     | #25       |
```

rfc/0054-sdk3-management-apis.md
chvck: Yup, will do when I squash the commits

rfc/0054-sdk3-management-apis.md
programmatix: That is true.  I think either way is reasonable.  Having them this way does make the create method nicely consistent with the update and delete methods.  But you're right, it does deviate from the BucketManager model - though I feel that in several ways we didn't nail the BucketManager API, and we can see this new CollectionManager API as an attempt to correct that and setup a better model for the future.
But really, I don't think I have a strong opinion either way.

avsej (approved):

rfc/0054-sdk3-management-apis.md
```diff
@@ -3597,13 +3659,48 @@ interface CollectionSpec {
...
+
+## UpdateCollectionSettings
+
+```
+interface UpdateCollectionSettings {
+    Duration MaxExpiry();
+    
+    Bool History();
+```
+
```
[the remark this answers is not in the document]
chvck: See above discussion between Dimitris and Graham.

rfc/0054-sdk3-management-apis.md
```diff
@@ -3569,6 +3618,19 @@ enum StorageBackend {
...
+* `HistoryRetentionCollectionDefault` (`bool`) - Whether to enable history retention on collections by default.
+  * URL query field: `historyRetentionCollectionDefault`
+  * If the SDK does not have the ability to differentiate between not set and `false` then a different type should be used instead, suggested:
+    * `enum HistoryRetentionCollectionDefaultSettings { ON, OFF }`
+
+* `HistoryRetentionBytes` (`int`) - The maximum size, in bytes, of the change history that is written to disk for all collections in this bucket.
+  * URL query field: `historyRetentionBytes`
+
+* `HistoryRetentionDuration` (`duration`) - The maximum number of seconds to be covered by the change history that is written to disk for all collections in this bucket.
+  * URL query field: `historyRetentionSeconds` sent as an int seconds value
```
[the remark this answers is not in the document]
chvck: If not set then send nothing.

rfc/0054-sdk3-management-apis.md
```diff
@@ -3068,7 +3073,49 @@ Nothing
 
 ### URI
 
-    POST http://localhost:8091/pools/default/buckets/<bucket>/collections/<scope_name> -d name=<collection_name> -d maxTTL=<maxExpiry>
+    POST http://localhost:8091/pools/default/buckets/<bucket>/scopes/<scope_name> -d name=<collection_name> -d maxTTL=<maxExpiry> -d history=<history>
```
programmatix, line 3076: Is this correct?  In JVM world we are using /pools/default/buckets/{bucketName}/scopes/{scopeName}/collections (which also matches the PATCH path below)

rfc/0054-sdk3-management-apis.md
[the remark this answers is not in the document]
jeffrymorris: That doesn't really clear things up :) if they are optional they should be part of the XxxOptions.

jeffrymorris (commented):
Still not convinced about XxxSettings.

PR #129, opened 2023-11-28
Clarify path level error handling for get projections

chvck:


rfc/0053-sdk3-crud.md
```diff
@@ -1268,6 +1268,8 @@ Given a JSON document with the following structure, the table below it shows the
 }
 ```
 
+Paths that return errors should be silently omitted, in the same way that a standard subdocument lookup operation would behave.
```
dnault, line 1271: I agree paths that return `PathNotFoundException` (and `PathMismatchException`?) should be silently omitted.

I'm less certain about errors that occur due to invalid arguments or invalid document state, like document too deep, path too long, etc.

dnault (approved):

avsej (approved):

brett19 (approved):

PR #130, opened 2023-12-11
Change type of QueryMetadata.signature() from JsonObject to JsonValue

dnault:
Since query signatures are a peripheral feature most users don't care about, it probably doesn't make sense to worry too much about changing existing SDKs. However, any new SDKs should support the full range of possible values.

### DIFF
--- a/rfc/0056-sdk3-query.md
+++ b/rfc/0056-sdk3-query.md
@@ -192,12 +192,14 @@ struct QueryMetaData {
     String RequestId();
     String ClientContextId();
     QueryStatus Status();
-    Optional<JsonObject> Signature();
+    Optional<JsonValue> Signature();
     List<QueryWarning> Warnings();
     QueryMetrics Metrics();
     Optional<JsonObject> Profile();
 }
 ```
+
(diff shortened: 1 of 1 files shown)


programmatix (dismissed):
It's a bit of a mess, but it's the API query apparently want at the REST level, so it doesn't give us a ton of options.

chvck (dismissed):

rfc/0056-sdk3-query.md
```diff
@@ -192,13 +192,15 @@ struct QueryMetaData {
     String RequestId();
     String ClientContextId();
     QueryStatus Status();
-    Optional<JsonObject> Signature();
+    Optional<AnyJsonValue> Signature();
```
brett19, line 195: nit: JsonValue is the name of the actual 'any value' json type.

dnault:
Changed `AnyJsonValue` to `JsonValue`, per Brett's comment.

chvck (approved):

brett19 (approved):

avsej (approved):

PR #131, opened 2023-12-18
Adding vector search

programmatix:


rfc/0052-sdk3-full-text-search.md
```diff
@@ -635,6 +635,7 @@ Important SDK considerations:
   * The `k` parameter is mandatory.  The FTS team suggest 3 is a sensible default to send from the SDKs.
 * There are new FTS index definition fields for defining the vector index, but they are inside the `plan` top-level field.  So there is nothing extra required from the SDKs to support these, as we already simply turn the untyped `plan` parameters into JSON.
 * The SDK (and Couchbase), at least in this first phase, have nothing to do with generating the vectors - both those stored in the documents, and those used for queries.  The user will be responsible for generating these, for example using the OpenAI Embeddings API.
+* The feature works unchanged with both global and scoped FTS indexes.
```
brett19, line 638: A few lines above this, it looks like a comment like this:

> (We believe: FTS PM to confirm.)

Was included in a previous commit which added Vector search.  I think we should probably get the answer to that and include the confirmed answer in these changes?

rfc/0052-sdk3-full-text-search.md
```diff
@@ -635,6 +635,7 @@ Important SDK considerations:
   * The `k` parameter is mandatory.  The FTS team suggest 3 is a sensible default to send from the SDKs.
```
brett19, line 635: If there are sensible defaults that could be sent, what is the reasoning behind them being required fields in the first place rather than having the server specify the defaults if they are not specified?  I think this is something we've been discussing in other contexts, where the SDKs should attempt to avoid specifying concrete 'default' values.

rfc/0052-sdk3-full-text-search.md
```diff
@@ -672,33 +675,29 @@ cluster.searchQuery("search_index_name", query);
 A more complex example showing multiple vector queries and all parameters:
 
 ```java
-// Setting all possible parameters on a VectorQuery.
+// Sending multiple `VectorQuery`s, and setting all possible parameters.
```
brett19, line 678: We support multiple queries through the use of ConjectionQuery and DisjunctionQuery.  Is there a reason behind exposing multiple vector queries through a different mechanism than what we have today?

I guess a related thing that comes to mind is that this API implies that a 'VectorQueries' object "wraps" a search object (ie: that the Vector Query is part of the search being performed).  But the actual API that's being exposed shows that the "query" and "knn" fields are both top-level concepts, which would make me think that the vector querying was meant to be exposed as a search option rather than as a SearchQuery type?  What was the reasoning behind hiding this distinction in the SDK API?

brett19:
Side-note: It looks like a part of this PR was accidentally committed directly into master, and part of it is in this PR's separate branch?  I can't seem to find the PR that committed the original part of the vector search changes that I commented on here.

rfc/0052-sdk3-full-text-search.md
programmatix: Yes that's in progress.  I've put up this very early draft just for initial feedback from the SDK PMs - it's not yet ready for review by the wider team.

rfc/0052-sdk3-full-text-search.md
programmatix: > We support multiple queries through the use of ConjectionQuery and DisjunctionQuery. Is there a reason behind exposing multiple vector queries through a different mechanism than what we have today?

Those two already have a defined meaning though - they combine queries inside the "query" field, e.g. the normal non-vector query.
They're also non-trivial and I don't want users to have to understand conjunctions vs disjunctions just to send a vector query.

> the vector querying was meant to be exposed as a search option rather than as a SearchQuery type?

Yes, it would be possible to expose in the options block instead - but that feels weird to me, and also that we're relegating vector search to fairly minor tertiary functionality.  Let's see what SDK PMs think.

programmatix:
> Side-note: It looks like a part of this PR was accidentally committed directly into master, and part of it is in this PR's separate branch? I can't seem to find the PR that committed the original part of the vector search changes that I commented on here.

Indeed, and I was hoping you could help me rectify this.
I intended to push the [previous commit](https://github.com/couchbaselabs/sdk-rfcs/commit/b491fcd663ca6f5987402cf4378337cf45140610) to my fork and then PR from there.  I pushed to this repo instead - and for reason, it allowed that (!).  Do you have permissions to just delete that commit from the history?

brett19:
I have removed the "Adding vector search" commit from master.

rfc/0052-sdk3-full-text-search.md
brett19: That's fair.  I guess ultimately the question I was trying to get at is:  _if_ the intent is that it is a 'top level' feature, such as the queries themselves.  Why isn't the FTS API modeled a different way?  Ultimately this is a net-new feature we are adding, and if we presuppose that PM is the one providing guidance on it's external interface, why the incongruence between teams?

programmatix:
Thanks @brett19.  This PR now replaced with the [fixed PR](https://github.com/couchbaselabs/sdk-rfcs/pull/132).

PR #132, opened 2023-12-20
Adding vector search

programmatix:


programmatix:
This is currently uploaded just for initial review by the SDK PM(s).  No review is requested from the wider team at this time.

programmatix:
Review from the wider SDK team is now welcome following initial PM review.

rfc/0052-sdk3-full-text-search.md
```diff
@@ -566,7 +566,250 @@ A query that matches nothing.
...
+So there is nothing extra required from the SDKs to support these, as we already simply turn the untyped `plan` parameters into JSON.
+* The SDK (and Couchbase in general), at least in this first phase, have nothing to do with generating the vectors - both those stored in the documents, and those used for queries.
+The user will be responsible for generating these, for example using the OpenAI Embeddings API.
+* The feature works with both global and scoped FTS indexes.
+
+SDK PM advises that the most common vector search use-cases are likely to be, in this order:
+
+* Performing a vector search in isolation (without a traditional FTS query).
+* A SQL++ statement is used to also perform a vector search.
+  * This is just passed in the string statement and is expected to return the same query results JSON as at present, so from the SDK POV it's just a passthrough with no SDK modifications needed.
```
programmatix, line 667: @jon-strabala can you confirm that the query results will be the same in this mode?

rfc/0052-sdk3-full-text-search.md
```diff
@@ -566,7 +566,275 @@ A query that matches nothing.
...
+It's mapped to the top-level `query` JSON field.
+* An individual vector query, which we can call a `VectorQuery`.  These map to the members of the new top-level `knn` array.
+It can contain options applicable just to that vector query such as "boost" and "k".
+* A way to perform multiple `VectorQuery`s, which we can call a `VectorSearch`.
+Maps to the top-level `knn` array.
+* Any options applicable to all forms of query (both the `SearchQuery` and the `VectorSearch`), which are in `SearchOptions` (skip, limit, explain, highlight, fields, etc.)  We can continue to use the existing `SearchOptions` object for this.
+* Any options applicable only to the `VectorSearch`, which we can call `VectorSearchOptions`.
+
+So the user will be performing a `SearchRequest` containing a `SearchQuery` and/or a `VectorSearch` and/or a `FutureSearchType`, optionally with some `SearchOptions` applying to all of those.  
+A `VectorSearch` comprises 1+ `VectorQuery`s and optionally some `VectorSearchOptions`.
```
programmatix, line 691: The distinction between "search" and "query" is pretty arbitrary here - there's quite a few concepts to name and limited ways of doing so.  Open to alternative naming schemes!

rfc/0052-sdk3-full-text-search.md
```diff
@@ -566,7 +566,275 @@ A query that matches nothing.
...
+
+Neither `VectorSearch` nor `VectorQuery` will extend the `SearchQuery` interface, which we will reserve for traditional FTS queries.
+
+Hence, none of this is compatible with the existing search interface `cluster.searchQuery(SearchQuery, SearchOptions)`.
+
+This takes us to a new API, `cluster.search(SearchRequest, SearchOptions)`, which has the flexibility to be able to perform traditional FTS and/or vector queries, and/or any future top-level queries.
```
brett19, line 697: This doesn't appear to expose the index name to search, but below in the examples it has the index name as the first parameter?

rfc/0052-sdk3-full-text-search.md
```diff
@@ -566,7 +566,275 @@ A query that matches nothing.
...
+* `uint32 numCandidates`.  Sent as a `k` number field in the JSON.  We name the field differently in the SDK for the same reason as `knn_operator`.  If not set by the user, FTS PM requests that the SDK send a default value of 3.  It controls how many results are returned and must be >= 1.  If < 1 the SDK will raise `InvalidArgument`.
+* `float boost`.  Sent as a `boost` number field in the JSON.  If not set, the field is not sent.
+
+`VectorQuery` does not extend the `SearchQuery` interface, which is now reserved for traditional FTS queries.
+
+### SearchRequest
+`SearchRequest` creation is platform-idiomatic.  There need to be two contruction options, allowing it to be created from either a `VectorSearch` or a `SearchQuery`.
+
+```
+SearchRequest.vectorSearch(VectorSearch vectorSearch, [SearchOptions searchOptions])
```
brett19, line 809: These methods appear to have a `SearchOptions` second parameter, but I think above the `SearchOptions` are indicated as being exposed as part of the `cluster.search()` / `scope.search()`?

rfc/0052-sdk3-full-text-search.md
```diff
@@ -566,7 +566,275 @@ A query that matches nothing.
...
+
+```
+SearchRequest.searchRequest([SearchQuery searchQuery], [VectorSearch vectorSearch], [SearchOptions searchOptions])
+```
+
+In this case, at least one of `SearchQuery` and `VectorSearch` must be provided and the SDK must raise `InvalidArgumentException` if not.
+
+`SearchRequest` does not extend the `SearchQuery` interface, which is now reserved for traditional FTS queries.
+
+### showrequest
```
brett19, line 829: Is this behaviour only part of the new `search()` methods, or should it be applied to the existing `searchQuery` method?  It says "for all queries", but I think it's only for the new method, which should be clarified.

rfc/0052-sdk3-full-text-search.md
```diff
@@ -566,7 +566,275 @@ A query that matches nothing.
...
+    OR
+}
+```
+
+`VectorSearch` does not extend the `SearchQuery` interface, which is now reserved for traditional FTS queries.
+
+### VectorQuery
+`VectorQuery` creation is platform-idiomatic:
+
+`VectorQuery.create(String vectorFieldName, float[] vector)`
```
brett19, line 789: In order to be more specific and support potential integrations with vector libraries, why not `VectorQuery.fromVector`?

rfc/0052-sdk3-full-text-search.md
```diff
@@ -566,7 +566,275 @@ A query that matches nothing.
...
+```
+
+### VectorSearch
+`VectorSearch` construction is platform-idiomatic:
+
+`VectorSearch.create(List<VectorQuery> vectorQueries, [VectorSearchOptions options])`
+
+The first parameter needs to be mandatory.
+
+`VectorSearch.create(VectorQuery vectorQuery, [VectorSearchOptions options])` can be added to make it easier to send a single `VectorQuery`.
```
brett19, line 770: Why does this accept VectorSearchOptions (which only has the operator, which would be unused with a single VectorQuery object)?  Relatedly, in the case of multiple queries, should the operator being used between them not be a more high-level object rather than being part of general options? For instance:
```
VectorSearch.fromMultiple(VectorQueryCombination, List<VectorQuery>)
VectorSearch.create(VectorQuery)
```

rfc/0052-sdk3-full-text-search.md
programmatix: Yes, that's a bug - it absolutely should have the search index name here, will fix.

rfc/0052-sdk3-full-text-search.md
programmatix: > Why does this accept VectorSearchOptions (which only has the operator, which would be unused with a single VectorQuery object)

It's forward-looking to when more options will be present in future, but yes, it's not useful right now.  Since discovering what a hard time some SDKs have with adding overloads later, I'm trying to anticipate these situations.
I don't have strong feelings on it - can certainly remove it from this optional sugar.

I'm not a big fan of the `fromMultiple` / VectorQueryCombination idea, I think, for similar reasons - there may well be future options.

rfc/0052-sdk3-full-text-search.md
programmatix: It's mentioned that construction is platform-idiomatic and this is intended to include the "VectorQuery.create(...)" naming.  If `new VectorQuery(...)`, or `VectorQuery.fromX(...)`, or some other form, is more idiomatic and better fits the SDK's existing FTS creation methods, it should use such.

rfc/0052-sdk3-full-text-search.md
programmatix: Yep you're right, this one's a bug and SearchOptions should not be here.

rfc/0052-sdk3-full-text-search.md
programmatix: That's correct, it applies to all queries from the new method.  I'll clarify it.

rfc/0052-sdk3-full-text-search.md
```diff
@@ -566,7 +566,250 @@ A query that matches nothing.
 JSON paths:
 * `match_none` = `null`: The JSON representation of a `MatchNoneQuery` is simply a `"match_none": null` entry in the query JSON Object.
 
-## Return Types
+## Vector search
+This is a feature being added to Couchbase Server 7.6 in the FTS service.
+
+The feature will initially be released in a Private Preview, and all SDK additions related to it should be annotated with the platform equivalent of @Stability.Volatile as changes may well be required following user feedback.
```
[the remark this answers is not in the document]
rajagp: Agree. Feature is definitely not private (it is public but could be be Beta or GA). Regardless as per our discussions, we should still mark APIs as volatile.

rfc/0052-sdk3-full-text-search.md
chvck: For SDKs which use them, is this float64 or float32?

programmatix: float32

rfc/0052-sdk3-full-text-search.md
```diff
@@ -566,7 +566,279 @@ A query that matches nothing.
...
+
+3. A combined search with both vector search and a traditional FTS query:
+
+```java
+float[] vector = OpenAI.createVectorFor("some query");
+
+SearchRequest request = SearchRequest
+        .searchQuery(SearchQuery.matchAll())
+        .vectorSearch(VectorSearch.create(VectorQuery.create("vector_field", vector)));
+
```
rajagp, line 744: Are we attached to `searchQuery` for historical reasons (as in what we have today so less ).It would be ideal to use` ftsSearch` and `ftsQuery` to align with `vectorSearch`/`vectorQuery` respectively- no?
A new developer reading the API without any of the context would find it hard to understand that searchQuery is actually FTS.

programmatix: Possibly - Ray and I have certainly wondered before why it's named "search" and not "fullTextSearch" or similar - but it potentially opens a rather large can of worms.  `.ftsSearch(SearchQuery.matchAll())` or `.fullTextSearch(SearchQuery.matchAll())` both look weirdly inconsistent to me - would probably want something instead like `.fullTextSearch(FullTextSearch[Query].matchAll())`, which looks cleaner.  But we have a lot of classes implementing that `SearchQuery` interface, so that's potentially an expensive addition - some SDKs may end up having to C&P a bunch.

rfc/0052-sdk3-full-text-search.md
```diff
@@ -566,7 +566,286 @@ A query that matches nothing.
...
+            vectorSearchOptions().vectorQueryCombination(VectorQueryCombination.AND));
+
+scope.search("search_index_name", request);
+cluster.search("search_index_name", request);
+```
+
+### VectorSearch
+`VectorSearch` construction is platform-idiomatic:
+
+`VectorSearch.create(List<VectorQuery> vectorQueries, [VectorSearchOptions options])`
```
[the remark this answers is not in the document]
programmatix: Ideally it would not use fluent-style anywhere: VectorQuery + VectorQueryOptions, and VectorSearch + VectorSearchOptions are the ideal, as they are more idiomatic to SDK3.

But, the Java SDK for some reason uses a fluent-style API for `SearchQuery`s, but an options block for SearchOptions.  Don't ask me why!  So I've tried to give enough latitude in the RFC for people to be able to implement something that ideally looks like an SDK3 API, but can instead fit the idiomatic style where needed.

rfc/0052-sdk3-full-text-search.md
```diff
@@ -566,7 +566,250 @@ A query that matches nothing.
...
+  "sort": [
+    "-_score"
+  ],
+  "limit": 5
+}
+```
+
+Important SDK considerations:
+
+* There is some uncertainty on whether it is mandatory to provide a `query` field containing a 'normal' FTS query.  This is being followed up with the FTS team.  As it makes sense to be able to perform solely a vector search, we will assume for now that will be the case.
```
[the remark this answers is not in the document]
programmatix: Raised this as [ticket](https://issues.couchbase.com/browse/MB-60312) as we urgently need a response to this.

rfc/0052-sdk3-full-text-search.md
vishaldhiman22: +1 to what @rajagp said.n It would be great to have consistency in function names ftsQuery, geoDistance, vectorQuery, etc.

programmatix: For the existing SearchQuery including geo I've used the existing names so the large corpus of existing SDK code based on it can be refused. Plus it is somewhat orthogonal to the vector search work, and I didn't want to unduly scope creep given the timescales required.

It's worth noting that this proposal was originally presented for PM review last year, and we are due to begin implementing tomorrow. I can still make small bug fixes but it is too late for feedback that will require such large ranging changes - unless everything is pushed back by a sprint. Let me know if you want to go that path; I'm happy to discuss renaming if given time to do it, though will need details on the desired names from PMs.

rfc/0052-sdk3-full-text-search.md
```diff
@@ -864,6 +874,21 @@ If the user has specified a vector search, the SDK should raise a `FeatureNotAva
...
+Iff a `VectorSearch` is included, before sending anything to the server, the SDK will check for the presence of this cluster capability which will only be returned by clusters that are fully upgraded to 7.6.0 or above:
+```
+  "clusterCapabilities": {
+   "search": ["vectorSearch"]
+  }
+```
+If it is not present the SDK will raise `FeatureNotAvailableException` with a message along the lines of "Vector queries are available from Couchbase Server 7.6.0 and above".
+
+The SDK will check for this capability in the most recent config it received.
+If it does not currently have one (in which case one should already be in the process of being asynchronously fetched), the SDK will wait for it to be available.
```
RiPont, line 888: I feel like this adds edge cases and potential performance problems for not much benefit.

If there is no config, chances are nothing is bootstrapped yet and the whole thing will fail anyways.

chvck (approved):

DemetrisChr (approved):

chvck:
@programmatix could you tidy the commit history a bit please?

dnault (approved):

PR #133, opened 2024-01-08
Update scope search index manager to include analyze document

chvck:


rfc/0054-sdk3-management-apis.md
```diff
@@ -1563,6 +1563,7 @@ The SDK should not try to intercept a "bucket.scope.index" name and extract "ind
 The implementation of `ScopeSearchIndexManager` is identical to `SearchIndexManagement`, except for the endpoints and some error handling details:
 
 * `GetIndex`, `UpsertIndex`, `DropIndex`: use endpoint `/api/bucket/{bucketName}/scope/{scopeName}/index/{indexName}` with GET, PUT and DELETE respectively.
+* `AnalyzeDocument`: use endpoint `/api/index/{bucketName}.{scopeName}.{indexName}/analyzeDoc` with POST, there is no scoped endpoint for analyze document.
```
programmatix, line 1566: Ah, really?  May have a bug in the Java SDK then.

chvck: Apparently so, from elsewhere (and from someone else)

> They told me:
so there isn’t a scoped endpoint for analyseDoc
you’ll have to provide the fully qualified index name

programmatix: This just seems to be a server bug - why have a scoped index endpoint for every operation except analyzeDoc?

DemetrisChr:
@programmatix Is the change suggested by this PR something we want to implement or do we want to still use the `/api/bucket/{bucketName}/scope/{scopeName}/index/{indexName}/analyzeDoc` endpoint which is not currently provided by FTS?

programmatix:
> @programmatix Is the change suggested by this PR something we want to implement or do we want to still use the `/api/bucket/{bucketName}/scope/{scopeName}/index/{indexName}/analyzeDoc` endpoint which is not currently provided by FTS?

Someone needs to raise it with FTS I think. It seems at the very least downright odd that only one endpoint is missing a scoped variant.

DemetrisChr:
> Someone needs to raise it with FTS I think. It seems at the very least downright odd that only one endpoint is missing a scoped variant.

I've filed an MB: https://issues.couchbase.com/browse/MB-60643

DemetrisChr:
The FTS team have now added a scoped endpoint for `AnalyzeDocument` (available from 7.6.2) so it should now be covered by the general rule:
> * For the other operations `PauseIngest` etc., the rule is: where the global index endpoint is `/api/index/{indexName}/ingestControl/pause`, the scoped index endpoint is `/api/bucket/{bucketName}/scope/{scopeName}/index/{indexName}/ingestControl/pause`.

I think we can now close this PR.

DemetrisChr:
Closing, there's now a scoped endpoint so this change is no longer necessary

PR #50, opened 2016-08-22
0001: Errata - Clarify on states and process.

daschl:
ping @ingenthr 

supersedes #23

PR #136, opened 2024-04-26
RFC-0078: Zone-Aware Replica Reads

avsej:


README.md
```diff
@@ -44,6 +44,7 @@ Coding happens all the time and is encouraged. We just recognize there is a poin
 | 64    | [SDK3 Field-Level Encryption](rfc/0064-sdk3-field-level-encryption.md) | David N.     | ACCEPTED |
 | 69    | [KV Error Map V2](rfc/0069-kv-error-map-v2.md)                         | Brett L.     | ACCEPTED |
 | 76    | [KV Subdoc Replica Reads](rfc/0076-subdoc-replica-reads.md)            | Charles D.   | ACCEPTED |
+| 77    | [Zone-Aware Replica Reads](rfc/0077-zone-aware-replica-reads.md)       | Sergey       | ACCEPTED |
```
programmatix, line 47: 77 is couchbase2 RFC (which should be in here, admittedly)

rfc/0077-zone-aware-replica-reads.md
```diff
@@ -0,0 +1,183 @@
...
+| Field          | Value                    |
+|----------------|--------------------------|
+| RFC Name       | Zone-Aware Replica Reads |
+| RFC ID         | 77                       |
+| Start Date     | 2024-04-25               |
+| Owner          | Sergey Avseyev           |
+| Current Status | DRAFT                    |
+| Revision       | #1                       |
+
+# Summary
```
programmatix, line 12: No GDoc?

rfc/0077-zone-aware-replica-reads.md
```diff
@@ -0,0 +1,183 @@
...
+(AZ).
+
+# Use Cases
+
+## No Preference
+
+This is current behaviour, where the SDK does not take into account server
+groups.
+
+![figure 1: No Prefrence](figures/0077-case-1-no-preference.svg)
```
programmatix, line 34: Any way to get this rendering in the diff?

rfc/0077-zone-aware-replica-reads.md
```diff
@@ -0,0 +1,183 @@
...
+
+## Local Only
+
+In this case, all reads will be done only from the local availability zone. It
+is the cheapest solution, although it reduces chances of getting data from
```
programmatix, line 39: increases chances surely?

rfc/0077-zone-aware-replica-reads.md
```diff
@@ -0,0 +1,183 @@
...
+```
+class LookupInAnyReplicaOptions {
+    // ...
+    ReadPreference readPreference { ReadPreference::NO_PREFERENCE };
+    // ...
+};
+```
+
+In all failure cases, when the SDK cannot handle operation, it must return `102
+DocumentUnretrievable` with human-readable explanation whenever it is possible
```
programmatix, line 102: One reason I was avoiding using the existing Get[Any|All]Replicas, is I consider their interfaces rather broken.  One concrete example - I have no easy way of disambiguating programmatically between the document not existing, and a network error.  There are other issues, including that GetAllReplicas swallows all substream errors, making it not a good building block for higher-level replica operations.

We need to consider that the feature needs to be added to transactions also - in fact this is the most important use-case since it is the one the requesting customer will use.  And I'm somewhat reluctant to add a broken interface into transactions.

rfc/0077-zone-aware-replica-reads.md
```diff
@@ -0,0 +1,183 @@
+# Meta
+
+| Field          | Value                    |
+|----------------|--------------------------|
```
programmatix, line 4: I don't see in this doc how the SDK exposes letting the user set which availability zone / server group the appserver is closest to?

I proposed "server group affinity" naming for this concept, as the feature is, at its most general, about sending to a particular server group rather than a local availability zone.
But I'll happily accept an argument that is too broad and that we should focus the feature and the naming on the first intended use-case.

rfc/0077-zone-aware-replica-reads.md
```diff
@@ -0,0 +1,183 @@
...
+cases:
+
+- number of replicas is 1, number of zones is 3. Neither active, nor replica is
+  in the local group.
+- number of replicas is 1, number of zones is 3. Only active node is in the
+  local group.
+- number of replicas is 1, number of zones is 3. Only replica node is in the
+  local group.
+
+Shall we warn users when locality restriction turns request into just GET,
```
programmatix, line 120: Yes, I raised the same concern in the one-pager - see there for Priya's thoughts.  

Warning seems a decent idea on the surface, but I worry it could get spammy.  I guess it only spams if there's a cluster misconfiguration, or in failover cases.

rfc/0077-zone-aware-replica-reads.md
```diff
@@ -0,0 +1,183 @@
...
+because only active node is in local group?
+
+## Q2
+
+Do we need to implement management API to create/retrieve server groups and
```
programmatix, line 125: Since this is so intimately tied to adding and removing nodes, and because the number of server groups is likely to be very low, I just don't see people managing them programmatically.

rfc/0077-zone-aware-replica-reads.md
```diff
@@ -0,0 +1,183 @@
...
+change node belonging to particular group?
+
+## Q3
+
+All currently supported server versions allow to configure server groups, but
+only 7.6.1 announce this information in configuration.
+[MB-60835](https://issues.couchbase.com/browse/MB-60835) How the SDK should
+behave with older server? Do we need to fail fast or silently fall back to
+current behavior?
```
programmatix, line 134: If we treat the desire to write to a specific server group as a request not a command, it gives us a lot of lee-way to handle situations like this.  If we don't have a config with server groups, we fallback to whatever the operation does normally.

rfc/0077-zone-aware-replica-reads.md
```diff
@@ -0,0 +1,183 @@
...
+};
+```
+
+Each of the operation should get new option, with default value of
+`ReadPreference::NO_PREFERENCE`.
+
+```
+class GetAllReplicasOptions {
+    // ...
+    ReadPreference readPreference { ReadPreference::NO_PREFERENCE };
```
programmatix, line 72: I'm not sure this has any meaning in Get*All*Replicas?

rfc/0077-zone-aware-replica-reads.md
```diff
@@ -0,0 +1,183 @@
...
+How the SDK should behave during rebalance? It cannot guarantee cheap traffic
+based on availability. Should we retry operations if the configuration says that
+replica has moved to different group?
+
+## Q7
+
+In case of more than two server groups configured, do we need to distinguish
+between non-local groups? Right now there is no weights associated with the
+groups, so we don't have enough information to reason about cost of
+communication with non-local groups.
```
programmatix, line 158: Not sure I understand this question.  The user tells us what server group they prefer to send traffic to from this SDK.   It doesn't matter how many other server groups there are - we're only interested in one.

rfc/0077-zone-aware-replica-reads.md
```diff
@@ -0,0 +1,183 @@
...
+
+## Q8
+
+Do we want to update GetResult and LookupInResult with the property, which would
+tell the user the name of the availability zone, where serving node has been
+deployed upon generating response?
```
programmatix, line 164: This could be a neat solution to a lot of the concerns above.  If we tell the user where we got the replica from, they could add their own warning if it doesn't match the server group they expected, rather than us needing to spam.

rfc/0077-zone-aware-replica-reads.md
programmatix: On transactions - let's focus on non-transactions for now though.  Once we nail that API it should be relatively straightforward to add it to transactions too.  I'm only bringing up transactions as an argument for why I'd maybe prefer we avoided using getAnyReplica() for this.

rfc/0077-zone-aware-replica-reads.md
```diff
@@ -0,0 +1,183 @@
...
+replicas.
+
+![figure 2: Local Only](figures/0077-case-2-local-only.svg)
+
+## Local First
+
+This case is a compromise between cost and availability, which trades speed. The
+idea is that the SDK will partition set of `[active, replica1, ...]` nodes into
+two groups: local and non-local. Then it will perform requests to local only,
+and only when neither of nodes will return data, send requests to non-local
```
programmatix, line 49: The trouble with LOCAL_FIRST is that if the non-local node doesn't return anything, it's usually going to be due to a timeout - network blip or undiscovered node failover.  So we only find out about it at the point the 2.5 KV user timeout has elapsed - and it's too late for us to try a fallback.

Instead of LOCAL_FIRST, we could support: The user can do an initial call with LOCAL_ONLY and a short timeout, and if that times out, try any fallback strategy they want.

README.md
programmatix: Oh I actually did add it before - https://github.com/couchbaselabs/sdk-rfcs/pull/135/files - it's just impossible to get 3 reviews for this repo..

avsej: I just followed procedure, and couchbase2 did not claim its number in README

rfc/0077-zone-aware-replica-reads.md
avsej: without google doc it will not require conversion later

rfc/0077-zone-aware-replica-reads.md
avsej: I also noticed, it renders in the doc, but neither absolute, nor relative (like this) is not getting picked up diff. Maybe only if I will make it full URL starting with `http://...`

rfc/0077-zone-aware-replica-reads.md
avsej: maybe I incorrectly stated it, but it is related to one of the questions later. Where by selecting string Local-Only strategy we might be in the situation, where selected group does not have any replica, and maybe not even master copy of the vbucket.

rfc/0077-zone-aware-replica-reads.md
avsej: new options are essentially change behavior, maybe we can agree that with this new option it will distinguish 
* network error 
* document is not found
* case where the selected strategy does not allow us to fulfill request
  * not enough local replicas
  * no replicas configured
  * even master is not local group

rfc/0077-zone-aware-replica-reads.md
avsej: I think the only way would be to ask the user to specify the name of the server group, just like it would appear in the configuration. I'm good with this name.

```
var options = new ClusterOptions();
options.serverGroupAffinity("Group 1");
```

We may even emit some warning if group is not in the config, or leaves the cluster.

README.md
programmatix: Well - it has been trying to claim it for 3 months...  It's just very hard to get things merged into this repo.

rfc/0077-zone-aware-replica-reads.md
programmatix: Yes that's it exactly.  The first version of this feature is expected to require the user to manually specify the server group.  Maybe in future we can somehow interrogate the environment to find out we're in AWS AZ1, and somehow know to map that to Server Group 1.

> We may even emit some warning if group is not in the config, or leaves the cluster.

Yes, maybe a good idea, as long as we're not spammy.  (Which I could see it being with 7.6, with the rate that pushes out configs)

rfc/0077-zone-aware-replica-reads.md
programmatix: If we want to move to a GDoc-less workflow, I think this support will be important.

rfc/0077-zone-aware-replica-reads.md
programmatix: Oh I see.  So it reduces chances of getting data, full stop.  I took it as meaning a replica instead of the active.

rfc/0077-zone-aware-replica-reads.md
programmatix: Yes, that seems a viable solution.

README.md
programmatix: Anyway, appreciate you've resolved this now, so resolving this conversation.

rfc/0077-zone-aware-replica-reads.md
programmatix: I may be changing my feeling on that..  I'm very reluctant to add a not-great interface into transactions.  

It is an option to only include getAnyReplica() in transactions, and only allow LOCAL_ONLY mode (with its better behaviour) - but perhaps not a great one.

rfc/0078-zone-aware-replica-reads.md
```diff
@@ -0,0 +1,183 @@
...
+DocumentUnretrievable` with human-readable explanation whenever it is possible
+that explains the details.
+
+# Open Questions
+
+## Q1
+
+What the SDK should do if the local group does not have any replica configured
+for the key, but the User request `Local-Only` strategy? Consider the following
+cases:
```
programmatix, line 111: Also the mixed-mode case, where the config may flip between providing server groups and not.

rfc/0077-zone-aware-replica-reads.md
avsej: Yes, but with google docs still optional (#50 still is not merged) :)

And in a way google docs would require to perform conversion losing all discussions behind.

rfc/0077-zone-aware-replica-reads.md
avsej: I believe after all it makes sense to add getAllReplicas and getAnyReplica into transactions and let them behave like regular operations, although implemented using lookupInAllReplicas and lookupInAnyReplica internally.

rfc/0077-zone-aware-replica-reads.md
avsej: Right, so for now -- no management API, and for QA we have cbdinocluster 0.0.43 that supports automatic groups provisioning from YAML cluster definition.

rfc/0077-zone-aware-replica-reads.md
avsej: I updated response to this question, basically the behavior for old servers is not defined as they do not support Zone-Aware-Replica-Reads in the way we describe it (which requires `serverGroup` markers in the configuration).

I think we should rather document this better to make it clear how to use it, and that whole cluster should have at least 7.6.2 on all nodes, and be balanced correctly.

rfc/0077-zone-aware-replica-reads.md
avsej: Right, also note that we always read from active, so if you have two replicas configured for the bucket, you might have two responses.

rfc/0077-zone-aware-replica-reads.md
avsej: This question was more about should be make any difference between other groups in terms of cost. And it makes sense to only care about Local Server Group.

rfc/0078-zone-aware-replica-reads.md
avsej: I added response from PRD that we do not support setups where some of the nodes are older than 7.6.2.

rfc/0077-zone-aware-replica-reads.md
avsej: I have added section for that. Lets leave the warning part as implementation detail.

avsej:
@programmatix I've uploaded update for transactions API. It documents `DocumentNotFound` exception.

rfc/0078-zone-aware-replica-reads.md
```diff
@@ -0,0 +1,346 @@
...
+the groups are not balanced properly on the cluster, or some nodes has been
+failed over. The previous strategy in this case would return
+`102 DocumentIrretrievable` error and refuse touching replicas from non-local
+group.
+
+Note that in this case we do not do second round trip after trying nodes from
+selected server group, but only use current topology to detect the Empty-Group
+case.
+
+![Selected Server Group or All Available](figures/0078-case-3-selected-server-group-or-all-available.svg)
```
programmatix, line 57: The inability to view these significantly impacts ability to review

rfc/0078-zone-aware-replica-reads.md
```diff
@@ -0,0 +1,346 @@
...
+
+## Selected Server Group
+
+In this case, all reads will be done only from the local availability zone. It
+is the cheapest solution, although it reduces chances of getting data from
+replicas (if the server groups are unbalanced).
+
+![Selected Server Group](figures/0078-case-2-local-only.svg)
+
+## Selected Server Group or All Available
```
programmatix, line 44: It's maybe worth considering as we think about any fallback options, that we perhaps want to recommend that the user  always has a fallback strategy in place:
```
try {
   return collection.getAnyReplica(docId, options().timeout("20ms").read_preference(SELECTED_SERVER_GROUP)
} catch DocumentUnretrievableException | DocumentNotFoundException {
  return collection.get(docId)
  // or collection.getAnyReplica() or collection.getAllReplicas()  - as they wish
}
```
This saves them from a bunch of edge cases - like a new doc not having been replicated, mixed-mode clusters, and possibly some rebalance & failover scenarios.  Working with replicas is inherently a bit flaky, and we should advise users how to protect themselves.

Plus they don't necessarily know if they are working with 7.6.2 or not.

So in that context, knowing the user is going to have a fallback strategy in place already - I do wonder if there's any benefit to having fallback modes like this?

rfc/0078-zone-aware-replica-reads.md
```diff
@@ -0,0 +1,346 @@
...
+## No Preference
+
+This is current behaviour, where the SDK does not take into account server
+groups.
+
+![No Prefrence](figures/0078-case-1-no-preference.svg)
+
+## Selected Server Group
+
+In this case, all reads will be done only from the local availability zone. It
```
programmatix, line 38: s/local availability zone/selected|preferred server group, I think?

rfc/0078-zone-aware-replica-reads.md
```diff
@@ -0,0 +1,346 @@
...
+Where `serverGroupName` matches name of the group as it seen in Admin UI of the
+server and in cluster configuration JSON.
+
+## Selecting Read Preference for Operations
+
+New enumeration should be defined for expressing different strategies for the
+read replica APIs:
+
+```
+enum ReadPreference {
```
programmatix, line 134: Slight nit:  SELECTED_SERVER_GROUP at least is a strict command rather than a preference.  Should this be ReadRequirement?

I don't feel strongly on this though.

rfc/0078-zone-aware-replica-reads.md
```diff
@@ -0,0 +1,346 @@
...
+
+Inside transaction closure, the replica read API takes simplified form, and
+always read from local group, which implies that the SDK will always throw `102
+DocumentIrretrievable` error for servers that do not expose `serverGroup`
+property in cluster configuration.
+
+```
+class TransactionAttemptContext {
+    // ...
+    public TransactionGetResult getReplica(Collection collection,
```
programmatix, line 190: Ah so maybe this has got lost in translation but I think this should be getReplicaFromPreferred[Server]Group() or similar.

(Or more accurately: I wish I had the time and capability to design a true getReplica() that handles all single-replica-read concerns.  But I cannot figure out what that should look like.  So until either inspiration strikes or I get more cycles to spend on it, the thought is to have a single-purpose method instead that is focussed on only this use-case.  That doesn't "salt the earth" from doing a true getReplica() at a later date.)

rfc/0078-zone-aware-replica-reads.md
```diff
@@ -0,0 +1,346 @@
...
+
+## Transactions
+
+Inside transaction closure, the replica read API takes simplified form, and
+always read from local group, which implies that the SDK will always throw `102
+DocumentIrretrievable` error for servers that do not expose `serverGroup`
+property in cluster configuration.
+
+```
+class TransactionAttemptContext {
```
programmatix, line 188: I feel a bit uneasy with transactions specs being outside of the hackmd docs where they are all grouped together.

Perhaps I can just copy this into the [API doc](https://hackmd.io/cHIcXWZSQOi23qpJJAa8OQ) once finalised?  We need a better long-term solution though.

rfc/0078-zone-aware-replica-reads.md
```diff
@@ -0,0 +1,346 @@
...
+* `101 DocumentNotFound`, when the server returns KV Code `0x01 ENOENT` for
+  *all* requests.
+
+* `102 DocumentIrretrievable`, when the SDK finds that there are nodes in local
+  group, or there is no group available with the name selected in connection
+  options.
+
+# Caveats
+
+The use must be aware of the following topics when dealing with
```
programmatix, line 217: typo: user

rfc/0078-zone-aware-replica-reads.md
```diff
@@ -0,0 +1,346 @@
...
+> All currently supported server versions allow to configure server groups, but
+> only 7.6.1 announce this information in configuration.
+> [MB-60835](https://issues.couchbase.com/browse/MB-60835) How the SDK should
+> behave with older server? Do we need to fail fast or silently fall back to
+> current behavior?
+
+The behavior of this feature with older servers (pre 7.6.2) is not defined. It
+is up to SDK to decide what to do (for example, it could return `102
+DocumentIrretrievable` in this case, because the configuration does not tell us
+anything about server groups assigned to the nodes)
```
programmatix, line 274: I feel we should nail this down.

What you've proposed makes sense to me, in the context of what I propose elsewhere about the user always have try-catch handling.

programmatix (changes requested):

rfc/0078-zone-aware-replica-reads.md
avsej: Make sense. I can update here with the reference to it.

rfc/0078-zone-aware-replica-reads.md
avsej: right

rfc/0078-zone-aware-replica-reads.md
avsej: next time I will either use PNG or try to figure out something else. To me the benefit of keeping history here and removing conversion step weigh more. 

Meanwhile with couple of clicks you can see it in the fork
https://github.com/avsej/sdk-rfcs/blob/0077-zone-aware-replica-reads/rfc/0078-zone-aware-replica-reads.md

I just figured that maybe it is because of the fork, it does not show, because it insert URLs of the images, it just cannot render them as they point to this repo, not to the fork. I think it is rather github bug.

image

rfc/0077-zone-aware-replica-reads.md
avsej: This strategy got simplified, the SDK does not do cascaded requests. It just filters the replica set according to the preference rules

rfc/0078-zone-aware-replica-reads.md
avsej: My though process was like this:

1. the connection options have it `preferredServerGroup` as the setting is global, and I want to avoid impression that it applies to *all* operations
2. then the enumeration describes different type of this preference, so technically later we can expand it with more sophisticated strategies like the one I described in the initial draft were we do cascaded requests: first try local group, and then if it gives us nothing -- do the regular request automatically without using try/catch block

rfc/0077-zone-aware-replica-reads.md
avsej: Okay, no warnings. There will be good documentation on how to prepare the cluster.

rfc/0078-zone-aware-replica-reads.md
programmatix: Ok, fair enough.  It's a good point that it aligns with the setting.

rfc/0078-zone-aware-replica-reads.md
avsej: I will update this with reference to the fallback example

rfc/0078-zone-aware-replica-reads.md
avsej: Updated with recent commits

rfc/0077-zone-aware-replica-reads.md
avsej: Lets keep first iteration simple, and add it later if we find real use-case or demand

programmatix (dismissed):

rfc/0078-zone-aware-replica-reads.md
```diff
@@ -0,0 +1,352 @@
...
+
+| Language    | Team Member    | Signoff Date | Revision |
+|-------------|----------------|--------------|----------|
+| .NET        |                | 2024-MM-DD   |          |
+| Go          |                | 2024-MM-DD   |          |
+| C/C++       |                | 2024-MM-DD   |          |
+| Node.js     |                | 2024-MM-DD   |          |
+| PHP         |                | 2024-MM-DD   |          |
+| Python      |                | 2024-MM-DD   |          |
+| Ruby        |                | 2024-MM-DD   |          |
```
avsej, line 349:
```suggestion
| Ruby        | Sergey Avseyev | 2024-06-10   | #1       |
```

rfc/0078-zone-aware-replica-reads.md
```diff
@@ -0,0 +1,352 @@
...
+## No Preference
+
+This is current behaviour, where the SDK does not take into account server
+groups.
+
+![No Prefrence](figures/0078-case-1-no-preference.svg)
+
+## Selected Server Group
+
+In this case, all reads will be done only from the selected server group. It
```
brett19, line 38: You should probably clarify here that there might be multiple nodes that satisfy the requirement of being in the selected server group with something like:
> reads will be done from all nodes in the selected server group

rfc/0078-zone-aware-replica-reads.md
```diff
@@ -0,0 +1,352 @@
...
+is the cheapest solution, although it reduces chances of getting data from
+replicas (if the server groups are unbalanced).
+
+![Selected Server Group](figures/0078-case-2-local-only.svg)
+
+## Selected Server Group or All Available
+
+This strategy allows using all available replicas, but only in case when the
```
brett19, line 46: Is this meant to mean "only in the case where the local server group doesn't exist" or "only in the case where the local server group has no nodes for the needed replica"

rfc/0078-zone-aware-replica-reads.md
avsej: Fixed.

rfc/0078-zone-aware-replica-reads.md
avsej: I have updated the text to explain that this third case is just a fallback from the third, to make sure that the SDK allows user to do it. Because in certain setups it is not possible to make group to cover all 1024 vBuckets (when number of copies is less than number of groups, e.g.: 1 active + 2 replicas < 4 groups).

rfc/0078-zone-aware-replica-reads.md
```diff
@@ -0,0 +1,359 @@
...
+    ReadPreference readPreference { ReadPreference::NO_PREFERENCE };
+    // ...
+};
+```
+
+In all failure cases, when the SDK cannot handle operation, it must return `102
+DocumentUnretrievable` with human-readable explanation whenever it is possible
+that explains the details.
+
+## Transactions
```
DemetrisChr, line 192: Do we want to formalize the transactions side of this RFC as an extension? (cc @programmatix)

rfc/0078-zone-aware-replica-reads.md
```diff
@@ -0,0 +1,359 @@
...
+    // ...
+    ClusterOptions preferredServerGroup(String serverGroupName);
+    // ...
+}
+```
+
+Where `serverGroupName` matches name of the group as it seen in Admin UI of the
+server and in cluster configuration JSON.
+
+## Selecting Read Preference for Operations
```
chvck, line 141: What should we do when the user specifies SELECTED_SERVER_GROUP on the op but doesn't specify a preferredServerGroup on ClusterOptions?

rfc/0078-zone-aware-replica-reads.md
```diff
@@ -0,0 +1,359 @@
...
+Having this information, the SDK will be able to filter list of the node
+indexes in `vBucketMap` to get local server group members.
+
+For example, lets say the configured SDK to use `"group_1"` as its local server
+group. Then if some key is mapped to vBucket `0`, the SDK can filter vector if
+indexes `[0, 1]` to only `[0, -1]` (using `-1` here for illustration), because
+server with index `1` belongs to `"group_2"`. In the result, the SDK only need
+to send `GET` request to retrieve active copy of the document.
+
+## Selecting Server Group for Connection
```
chvck, line 125: What should we do when the specified server group just doesn't exist at all in the config?

avsej: We cannot know for sure why it does not exist. So we still assume that missing group is some temporary event, and try to filter the list using it. In other words, the operation will return `102 DocumentUnretrievable` because the result set is empty.

rfc/0078-zone-aware-replica-reads.md
avsej: Same behaviour, the SDK would try to use empty group name to filter the nodes, and because none of the nodes belongs to empty group -- the operation will fail with `102 DocumentUnretrievable`

rfc/0078-zone-aware-replica-reads.md
```diff
@@ -0,0 +1,369 @@
...
+
+## Selecting Read Preference for Operations
+
+New enumeration should be defined for expressing different strategies for the
+read replica APIs:
+
+```
+enum ReadPreference {
+    NO_PREFERENCE,
+    SELECTED_SERVER_GROUP,
```
dnault, line 153: Should this be `PREFERRED_SERVER_GROUP` to match the name of the cluster option?

rfc/0078-zone-aware-replica-reads.md
```diff
@@ -0,0 +1,369 @@
...
+
+## Selecting Server Group for Connection
+
+To allow user to select preferred server groups, new setter/property for
+`ClusterOptions` must be introduced:
+
+```
+class ClusterOptions {
+    // ...
+    ClusterOptions preferredServerGroup(String serverGroupName);
```
dnault, line 133: Should we have a standard name for a corresponding connection string property?

If so, would that nudge us towards calling this just `serverGroup` which is ~50% easier to type? :-)

rfc/0078-zone-aware-replica-reads.md
mikereiche: Wouldn't a user have a better chance of figuring out "read_preference option specified, but cluster does not have a preferredServerGroup" versus DocumentUnretrievable?

Edit: I guess the explanation can go in the "human-readable explanation whenever it is possible
that explains the details"

rfc/0078-zone-aware-replica-reads.md
mikereiche: Kind of of what Charlie already asked, but - Is there anyway to detect the cluster having group_1, group_2, and the SDK being configured with say group_one (i.e. a typo) ?  Does the configuration have a list of server group names independent of the server group name paired with hostname?  It's almost as though I don't trust users...

mikereiche (dismissed):

rfc/0078-zone-aware-replica-reads.md
```diff
@@ -0,0 +1,369 @@
...
+In general, the user must be expecting that the document will not be available
+in local group, so
+
+```
+try {
+   return collection.getAnyReplica(docId,
+                                   options()
+                                    .timeout("20ms")
+                                    .read_preference(SELECTED_SERVER_GROUP))
+} catch DocumentUnretrievableException | DocumentNotFoundException {
```
programmatix, line 65: @avsej for my understanding - I thought the replica methods would never raise DocumentNotFoundException?

programmatix (dismissed):

rfc/0078-zone-aware-replica-reads.md
```diff
@@ -0,0 +1,369 @@
...
+
+## Selected Server Group
+
+In this case, all reads will be done from all nodes in the selected server group. It
+is the cheapest solution, although it reduces chances of getting data from
+replicas (if the server groups are unbalanced).
+
+![Selected Server Group](figures/0078-case-2-local-only.svg)
+
+## Selected Server Group with Fallback
```
programmatix, line 44: @avsej this isn't in the ReadPreference enum below - which one is correct?

rfc/0078-zone-aware-replica-reads.md
```diff
@@ -0,0 +1,369 @@
...
+```
+
+The method might throw the following errors:
+
+* `101 DocumentNotFound`, when the server returns KV Code `0x01 ENOENT` for
+  *all* requests.
+
+* `102 DocumentUnretrievable`, when the SDK finds that there are nodes in local
+  group, or there is no group available with the name selected in connection
+  options.
```
DemetrisChr, line 230: This might be inconsistent with the non-transactional `GetAnyReplica` behaviour. The [error handling RFC](https://github.com/couchbaselabs/sdk-rfcs/blob/master/rfc/0058-error-handling.md#102-documentunretrievable) suggests that `DocumentUnretrievable` is the error to throw (not `DocumentNotFound` as suggested here) when all `GetReplica` requests fail with `0x01 ENOENT`:
> ### 102 DocumentUnretrievable
> Raised When
> *  In getAnyReplica, the getAllReplicas returns an empty stream because all the individual errors are dropped (i.e. all returned a DocumentNotFound)

rfc/0078-zone-aware-replica-reads.md
programmatix: Discussed with Sergey on Slack, and this option is either removed or optional.

rfc/0078-zone-aware-replica-reads.md
```diff
@@ -0,0 +1,369 @@
...
+
+```
+class TransactionAttemptContext {
+  // ...
+  public TransactionGetResult
+  getReplicaFromPreferredServerGroup(Collection collection, String id,
+                                     TransactionGetReplicaOptions options);
+  // ...
+}
+```
```
emilienbev, line 212: It would be nice to link the to Transactions API spec RFC, to make it easier to find the spec:

- https://github.com/couchbaselabs/couchbase-transactions-specs/blob/master/transactions-api.md#getreplicafrompreferredservergroup
- https://github.com/couchbaselabs/couchbase-transactions-specs/blob/master/transactions-extensions.md#extreplicafrompreferredgroup

rfc/0078-zone-aware-replica-reads.md
avsej: At this stage we have released some SDKs with `SELECTED_SERVER_GROUP`. I am not sure that all platform allow aliases for enum values.

So lets keep it as is, and with major update we might rename it

rfc/0078-zone-aware-replica-reads.md
avsej: Agree. Updated

rfc/0078-zone-aware-replica-reads.md
avsej: Discussed with @programmatix this option have to be mandatory for the implementation (although might not be part of the Public API), because it is necessary for [ExtGetMulti](https://github.com/couchbaselabs/couchbase-transactions-specs/pull/2) extension of the transactions.

rfc/0078-zone-aware-replica-reads.md
avsej: Agree, will fix it

rfc/0078-zone-aware-replica-reads.md
avsej: Done

programmatix (approved):

PR #138, opened 2024-05-22
Add base64 vector search

chvck:


programmatix (approved):

dnault (approved):

rfc/0052-sdk3-full-text-search.md
```diff
@@ -809,11 +809,27 @@ enum VectorQueryCombination {
...
+Both parameters in either signature need to be mandatory.
+
+Base64 vector queries are only supported from server version 7.6.2, however there is no mechanism to check if the feature is supported.
+Usage on a version (after 7.6.0 and) prior to 7.6.2 will yield a 400 status code from the search service containing error text like:
+```
+rest_index: Query, indexName: global_vector_index_989008, err: bleve: QueryBleve parsing searchRequest, err: k must be greater than 0 and vector must be non-empty
+```
+For this a CouchbaseException will be raised as usual.
+
+Converting the base64 encoding to float[] on the client-side was considered and rejected due to the partial intention of base64 encoding being to reduce network usage.
```
ingenthr, line 823: This is a bit ambiguious.  Do you mean converting at the client for older versions was considered, or more in general?

chvck: > Do you mean converting at the client for older versions was considered

This one, converting at the client so that what we send is supported by all server versions that support vector search.

chvck:
I'm merging to get this in, if we need an amendment to some wording then happy to do that as another PR.

PR #141, opened 2024-11-11
RFC-0047: Unified User Agent

avsej:


rfc/0047-unified-user-agent.md
```diff
@@ -0,0 +1,166 @@
...
+* Required
+
+The identifier allows to uniquely distinguish between the different SDKs used.
+The following identifiers must be followed by the implementations:
+
+| SDK           | Identifier |
+| ------------- | ---------- |
+| .NET          | `dotnet`   |
+| C++           | `cxx`      |
+| Go            | `golang`   |
```
DemetrisChr, line 54: @chvck Should we split this to `gocb` and `gocbcore` which more closely aligns with what gocb/gocbcore currently do?

dnault (dismissed):

rfc/0047-unified-user-agent.md
DemetrisChr:
```suggestion
| Go            | `gocb`     |
| Go Core       | `gocbcore` |
```

rfc/0047-unified-user-agent.md
```diff
@@ -0,0 +1,166 @@
...
+| Libcouchbase  | `lcb`      |
+| Node.js       | `nodejs`   |
+| PHP           | `php`      |
+| Python        | `python`   |
+| Ruby          | `ruby`     |
+| Rust          | `rust`     |
+| Scala         | `scala`    |
+| Kafka         | `kafka`    |
+| ElasticSearch | `es`       |
+
```
thejcfactor, line 67:
```suggestion
| Java Columnar      | `couchbase-java-columnar` |
| Node.js Columnar   | `nodejs-columnar`         |
| Python Columnar    | `python-columnar`         |

```

Seems like we might as well add the Columnar SDKs since we are making an update when they are already in existence.

ingenthr:
I don't have a strong opinion, but just to voice it… 

Seeing things like "java/3.7.5" in the string seem a bit misleading.  It's not really "java".  I know not many unfamiliar with Couchbase will be looking at this, but it does make be wonder if a couple more characters like "javacb" or "cbjava" disambiguate it better.  There may be some collisions that could be confusing.  "python/4.1.3" could collide next year with an actual python runtime.  

If ya'll think I'm being too pedantic, I'm good to go with the suggestions.

Also, regarding "couchbase-java-columnar" (23 chars?), the server side is going to use only up to the first 32 characters or a space, whichever comes first, in statistics aggregation (but not logging).  so something like "javacbcol" might be better to avoid a git hash or the like being excluded.

avsej:
@DemetrisChr, @thejcfactor, what do you think about Columnar Identifiers? Could you make code suggestions and we will be able to merge this RFC?

dnault (dismissed):

avsej:
> Also, regarding "couchbase-java-columnar" (23 chars?), the server side is going to use only up to the first 32 characters or a space, whichever comes first, in statistics aggregation (but not logging). so something like "javacbcol" might be better to avoid a git hash or the like being excluded.

@ingenthr it does not look like kv_engine tracks only first 23 chars

According to the code, it discards everything after first space character.

https://github.com/couchbase/kv_engine/blob/93f8d409cd7e12d876fabe1c33e54f1f5905a58d/daemon/sdk_connection_manager.cc#L21-L49

programmatix (dismissed):

rfc/0047-unified-user-agent.md
```diff
@@ -0,0 +1,187 @@
...
+| Python        | `python`   |
+| Ruby          | `ruby`     |
+| Rust          | `rust`     |
+| Scala         | `scala`    |
+| Kafka         | `kafka`    |
+| ElasticSearch | `es`       |
+| Java Columnar      | `couchbase-java-columnar` |
+| Node.js Columnar   | `nodejs-columnar`         |
+| Python Columnar    | `python-columnar`         |
+
```
DemetrisChr, line 71:
```suggestion
| Go Columnar        | `gocb-columnar`           |

```

rfc/0047-unified-user-agent.md
```diff
@@ -0,0 +1,188 @@
...
+# Signoff
+
+| Language   | Team Member            | Signoff Date   | Revision |
+|------------|------------------------|----------------|----------|
+| .NET       |                        |                |          |
+| Go         |                        |                |          |
+| Java       |                        |                |          |
+| Kotlin     |                        |                |          |
+| Node.js    |                        |                |          |
```
thejcfactor, line 181:
```suggestion
| Node.js    |  Jared Casey                      |   2025-04-03             |  1        |
```

thejcfactor (approved):
LGTM 👍

dnault (approved):

davidkelly (commented):
👍

davidkelly (approved):

PR #99, opened 2020-10-14
Update SDK 3 FTS to further detail facets in results

chvck:

Issue #13, opened 2016-01-13
0007_cluster_level_auth


brett19:
This issue needs to cover the transition of many of our services to potentially being cluster-level rather than the current paradigm of most operations hanging off our bucket objects.  This specifically relates to the transition of our authentication model away from management/bucket specific to being more generic to a cluster.

dodilp:
Brett,

This caught my eye -- can we chat more about this?

Thanks,
Don

brett19:
@jeffrymorris @mnunberg @daschl @simonbasle @ToddGreenstein
Quick example I came up with, I'll put this into the RFC shortly.
https://gist.github.com/brett19/1a2002ae5ad7270d8779

brett19:
As per @jeffrymorris, the credential provider is more likely correctly called 'BucketCredentials' or something similar rather than 'BucketAuthenticator'.

brett19:
This has moved to Google Docs:
https://docs.google.com/document/d/1CD5OL1ez7euCiLJT91zdWY9R4tW_bkGZ0wsFk1UDtyY/edit

daschl:
@brett19 this thing has been signed off I think by everyone, we should move it to accepted I guess. Do you want to convert it to markdown and move it here? If not I can do it too.

daschl:
opened a separate ticket for conversion, closing.

PR #146, opened 2025-04-02
CBD-6254: Add prefilter option to Vector Search

thejcfactor:


dnault (approved):

avsej (dismissed):

avsej:
Does it make sense to reset revision column in the Sign-Off table, so that everybody would re-sign-off the RFC?

thejcfactor:
Updated the changelog.

Also, w.r.t. the Sign-Off table...I suppose that would make since to see if we can get it refreshed w/ an updated revision.

avsej:
A lot of people signed off revision `#7`, while this one would be `#13`

dnault (approved):

thejcfactor:
I wonder if the approach can be making sure we have FIT tests to validate up to the current revision?  Then if all SDKs are green that can serve as the appropriate indication to update the sign-off table.

I will look into #99 and see what updates need to be made (and also see what I might add to FIT) 👍

programmatix (approved):
I figured this would either be a trivial feature, or one that turned out to be unexpectedly knotty - and it would purely depend on whether the entirety of the SearchQuery API could be safely embedded in this way.  Turns out it can, and it's the former case, and I'm relieved :) 

The one thin

avsej (approved):

PR #151, opened 2025-08-21
Update 0048-sdk3-bootstrapping.md for credential rotation

dnault:
Add a section describing a proposed new ~`DynamicAuthenticator`~ `DelegatingAuthenticator` for periodically refreshing credentials.

UPDATE: The proposal has evolved; current version is to add a new `Cluster.setAuthenticator` method that lets a user replace the current authenticator with a new one.

### DIFF
--- a/rfc/0048-sdk3-bootstrapping.md
+++ b/rfc/0048-sdk3-bootstrapping.md
@@ -200,7 +200,12 @@ A user should be able to supply a new Couchbase credential without having to res
 To support this, the SDK should allow replacing the Authenticator used by a `Cluster`.
 The recommended way to support this is for `Cluster` to have an instance method called `setAuthenticator` that takes the new authenticator as an argument.
 
-SDK implementers must take care to ensure it's safe to call `setAuthenticator` at any time, from any thread, and concurrently with an authentication operation.
+The `setAuthenticator` method should do the same validation as `Cluster.connect()`.
+Specifically, if the new authenticator requires TLS but TLS is not enabled, `setAuthenticator` should throw `InvalidArgumentException` or the idiomatic equivalent.
+
+The API reference documentation for `setAuthentictor` should clearly state that setting a new authenticator does not change the authentication status of existing connections.
+
+SDK implementers must ensure it is safe to call `setAuthenticator` from any thread at any time, including concurrently with an authentication operation.
 
 # Changelog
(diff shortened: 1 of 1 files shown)


brett19:
Interesting approach.  Have you given any consideration to the fact that Authenticators are public interfaces in most SDKs, but the methods you invoke are typically private (at least, this was the original Authenticator design).

dnault:
> Interesting approach. Have you given any consideration to the fact that Authenticators are public interfaces in most SDKs, but the methods invoked by the SDK within it are typically private (at least, this was the original Authenticator design).

@brett19 That's my understanding, yeah. `DynamicAuthenticator` would be part of the SDK, so it's okay for it to delegate to the "internal" methods of the standard authenticator provided by the user (which is either a CertificateAuthenticator or a PasswordAuthenticator).

I feel like there's a really good change I haven't understood your concern, though. Can you elaborate, please?

dnault:
> the methods invoked by the SDK within it are typically private (at least, this was the original Authenticator design).

@brett19 Oh wait, are you saying that in some SDK languages, the decorator pattern won't work because `DynamicAuthenticator` won't have access to the internal methods of the delegate authenticator?

brett19:
In hindsight, I miss-understood the premise of this when I first read it.  I forgot that users would be able to use the typical Authenticator's which are already exposed.  What was the reason for using a callback approach with a cache lifetime as opposed to just allowing the user to use a "SetAuthenticator" method on the DynamicAuthenticator as needed/wanted?

brett19:
You previously had clarifications related to the safety requirements in terms of concurrent access, you may want to keep those kinds of warnings in the context of the DelegatingAuthenticator, since there could be calls being made to it concurrently while the user updates it.  Lgtm though!

brett19 (approved):

dnault:
Let's use the approach Dimitris pioneered: don't worry about suggesting a change to the sign-off block. PR approval will count as sign-off, and we can update the sign-off block all at once in a separate commit so we don't lose individual PR approvals.

thejcfactor (approved):

mikereiche (approved):

mikereiche (approved):

thejcfactor (approved):

emilienbev (approved):

PR #152, opened 2025-08-26
RFC-55: Clarify that provided transcoders should not validate common flags

DemetrisChr:


DemetrisChr:
I will update the sign-off table based on any approvals after this is merged with a separate commit (to avoid GitHub discarding reviews when making the changes in this PR)

PR #153, opened 2025-09-03
RFC-67: Extended SDK Observability Revision 26

DemetrisChr:
This revision includes updates to bring our attributes consistent with the [OpenTelemetry Semantic Conventions v1.37](https://github.com/open-telemetry/semantic-conventions/tree/v1.37.0)


rfc/0067-extended-sdk-observability-rto-v2.md
```diff
@@ -0,0 +1,1144 @@
...
+| Server port | net.peer.port | server.port | |
+| Peer address | Did not exist | network.peer.address | |
+| Peer port | Did not exist | network.peer.port | |
+
+**Notes:**
+
+1. In v0, one of the possible values was "Success". In v1, the attribute should be omitted if there was no error.
+2. In v0, the value was "TCP.IP". In v1, it should be "tcp" instead.
+
+Existing implementations should not change the conventions that they emit by default. `OpenTelemetryRequestTracer` and `OpenTelemetryRequestMeter` should recognise the `OTEL_SEMCONV_STABILITY_OPT_IN` and emit the new stable conventions on an opt-in basis, as outlined in the [OpenTelemetry Database Semantic Conventions v1.37 document](https://github.com/open-telemetry/semantic-conventions/blob/v1.37.0/docs/database/database-spans.md):
```
DemetrisChr, line 1000: I'm undecided if this is the right approach. Should we support sending the new semantic conventions to the `Meter`/`Tracer` interfaces or only implement the conversion to the new conventions within the OpenTelemetry-specific concrete implementations?

Having it all in the OpenTelemetry-specific implementations means that the SDK is agnostic to OpenTelemetry conventions. However, it could cause confusion if the attributes we call the interfaces with attributes that have names that are Otel-like, but are not quite following the current Otel spec. Users implementing the interfaces on their own might also want to get the new semantic conventions without having to do extra work on their side.

rfc/0067-extended-sdk-observability-rto-v2.md
```diff
@@ -0,0 +1,1144 @@
...
+
+This attribute is a standard OpenTelemetry attribute and should be placed on all spans to uniquely identify them for couchbase. The attribute key is "`db.system`" and the value is `"couchbase"`.
+
+## Observability Metrics
+
+The Meter interface is the main abstraction and central place to handle metrics information. Note that while it is not an internal class, it is not one that a normal user would use directly - but instead use either the default implementation or adapters we ship. Advanced users can build their own implementations of the interface though.
+
+```
+interface Meter {
+     ValueRecorder valueRecorder(String name, Map<String, String> tags)
```
DemetrisChr, line 705: ValueRecorders no longer exist in Otel, they only existed in the v0.x versions. There are a number of more specific [instruments](https://opentelemetry.io/docs/concepts/signals/metrics/#metric-instruments) (counters, gauges) in the v1 spec, but our API is very inflexible at the moment, adding methods to interfaces would be a breaking change in some SDKs, so I think it's probably best to not introduce the new instruments unless we have any new metrics requiring them.

Though we need to consider how we would evolve the Meter API. Even with a MeterV2 could end up in a situation if a new type of instrument is added again. Additions to the API are not breaking from OpenTelemetry's point of view because their API is not an interface.

One option is to use the OpenTelemetry APIs within the SDK directly, instead of using interfaces. Users can register a global MeterProvider (or TracerProvider), and the SDK can access the Meter internally from it, without exposing any OpenTelemetry API.

rfc/0067-extended-sdk-observability-rto-v2.md
DemetrisChr: I think I'm leaning towards changing this to say that the new semantic conventions should be implemented within the SDK, that means all implementations of the `RequestTracer`, `Meter` interfaces can benefit & it's more consistent with the fact that it's the SDK itself implementing the conventions, and the `OpenTelemetryRequestTracer` and `OpenTelemetryRequestMeter` only relay the metrics/spans to an Otel-compatible backend.

dnault (dismissed):
Thanks for summarizing all the changes in that compatibility section!

rfc/0067-extended-sdk-observability-rto-v2.md
```diff
@@ -596,90 +605,104 @@ This attribute is a standard OpenTelemetry attribute and should be placed on N1Q
 | Views | Not present |
 | Analytics | The actual analytics query |
 
+Sanitization should be performed on the statement to exclude sensitive information, by replacing all literals (e.g. String, Numeric, Boolean, Date and Time) with a placeholder, `?`.
```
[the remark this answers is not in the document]
chvck: If we're going to do sanitization can you please include a regex or some pseudocode of how that should work? Otherwise we're going to end up with SDKs doing different things, and having different bugs.

chvck, line 608: If we're going to do sanitization can you please include a regex or some pseudocode of how that should work? Otherwise we're going to end up with SDKs doing different things, and having different bugs.

rfc/0067-extended-sdk-observability-rto-v2.md
```diff
@@ -505,51 +511,54 @@ Note that most APMs and OpenTelemetry consumers have fairly primitive indexing c
...
+| Analytics | `analytics` |
+| Management ("ns\_server" / 8091\) | `management` |
+| Transactions | `transactions` |
+| Eventing | `eventing` |
 
-Implementation note: the management service should only be used by requests which actually go to the cluster manager. If a management operation i.e. executes a N1QL query it should use the query service instead. The span name already reflects the actual operation type.
+**Implementation notes:**
+
+  * The management service should only be used by requests which actually go to the cluster manager. If a management operation i.e. executes a N1QL query it should use the query service instead. The span name already reflects the actual operation type.
+  * For SDK API operations which need to perform multiple actual underlying operations, where these operations are performed against a number of different services (e.g. `Ping`), the service identifier should be omitted.
```
programmatix, line 540: Well, except transactions, which has an explicit service type.

rfc/0067-extended-sdk-observability-rto-v2.md
programmatix: Yes, this feels a bad path.  The SDKs do not parse the query string, it's too complex.

We could only include the statement if the user provided 1 or more named or positional parameters?  That captures the spirit of what Otel says:

> Parameterized query text SHOULD NOT be sanitized. Even though parameterized query text can potentially have sensitive data, by using a parameterized query the user is giving a strong signal that any sensitive data will be passed as parameter values, and the benefit to observability of capturing the static part of the query text by default outweighs the risk.

rfc/0067-extended-sdk-observability-rto-v2.md
programmatix: > unless we have any new metrics requiring them

N.b. ExtObservability for transactions does require standard counters.

But I agree with your comments, both that our Meter interface is no longer suitable, and also that it's a challenging change, and one we don't need to tackle immediately.  I suspect we need a MeterV2 that mirrors current Otel API and will just have to accept that there may be a MeterV3 one day also.

rfc/0067-extended-sdk-observability-rto-v2.md
```diff
@@ -0,0 +1,1144 @@
...
+* **Type:** ValueRecorder
+* **Value:** Response Latency in *microseconds  which covers the complete request / response cycle (think: "outer request span latency", covering encode, dispatch, retries etc). Note: for bulk operations, each individual request counts individually.*
+* **Name:** "`db.client.operation.duration`"
+* **Instrumented at:** when the response has been decoded, before it is marked as completed towards the user
+* **Tags:**
+  * "`db.couchbase.service`": Same as in [attribute service identifier](#attribute:-system)
+  * "`db.operation.name`": Same as the "outer span names" - for example `"get"`, `"get_and_lock"` etc.
+  * "`db.namespace`": The bucket name
+  * "`db.couchbase.scope`". The scope name.
+  * "`db.couchbase.collection`".  The collection name
```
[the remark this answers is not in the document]
programmatix: I feel we should definitely keep the separate bucket, scope and collection attributes.  It means users can easily find metrics e.g. across a particular scope without having to do regex wrangling or similar.  And IME OTel consumers tend to be very basic on queries, and may not even support splitting attribute values.

Whether we duplicate things by also reporting db.collection.name, I don't have a strong feeling on.  My gut says no - better to DRY, we can add it later if needed, and there is of course a network cost to every attribute esp longer ones like this.

rfc/0067-extended-sdk-observability-rto-v2.md
```diff
@@ -0,0 +1,1144 @@
...
+| :---- | :---- | :---- | :---- |
+| KV | Present | Present | Present |
+| `scope.query()` `scope.search()` | Present | Present | Absent |
+| `cluster.query()` `cluster.search()` | Absent | Absent | Absent |
+| `cluster.transactions().run(...)` | Absent | Absent | Absent |
+
+From a high-level SDKs (libcouchbase) perspective when working with a low-level SDK (such as Node.js), we can simply have the higher level SDK ignore the operations metric being generated by the lower level SDK, and have it generate its own metric for that instead.
+
+> [!NOTE]
+> Metric values for `db.client.operation.duration` are reported to the SDK's `ValueRecorder` interface as integers representing microseconds. These should be converted to _seconds_ when reported to the OpenTelemetry `Histogram`.
```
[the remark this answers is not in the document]
programmatix: Hmm.  Regardless of what Otel says, I am concerned this is going to be bad UX.  A simple read could take under 100 micros, easily.  Users will be squinting at 0.000078 and trying to figure out what that means in real terms.  I feel OTel have just got it wrong on this one.

(Admittedly on the flip side, it means that on timeout they are looking at a friendlier 2.5 rather than 2500000.  But those should be rarer cases.)

This is also one of the times where the abstraction presents trickiness.

Prometheus, which is the real defacto metric standard, says the metric name should contain the unit, and the values themselves are unitless.  That doesn't really mesh with what OTel wants.  (Though Prometheus does also suggest that seconds should be the base unit for time - https://prometheus.io/docs/practices/naming/#base-units)

rfc/0067-extended-sdk-observability-rto-v2.md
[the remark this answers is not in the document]
DemetrisChr: I agree, I think let's just leave it as is for now, without `db.collection.name`.

rfc/0067-extended-sdk-observability-rto-v2.md
DemetrisChr: > N.b. ExtObservability for transactions does require standard counters.

I think we're lucky that the Meter interface already has counters in Go, which will help with this. In most other languages it's possibly easier to extend the Meter interface without making a breaking change (in a similar way to Java's default methods).

rfc/0067-extended-sdk-observability-rto-v2.md
DemetrisChr: Updated based on the above.

rfc/0067-extended-sdk-observability-rto-v2.md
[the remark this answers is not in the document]
DemetrisChr: Updated to say that the statement should only be included if named/positional parameters are used, as we don't want to parse the statement.

> This also means that db.operation.name will have to change so it's always present, if we are not always including the statement.
I think this might not be necessary – the top-level span has a name.

DemetrisChr: Updated to say that the statement should only be included if named/positional parameters are used instead.

dnault (dismissed):

avsej (dismissed):

rfc/0067-extended-sdk-observability-rto-v2.md
```diff
@@ -592,7 +596,7 @@ Should be placed on all operations which are at the collection level. The attrib
 
 #### Attribute: Statement
 
-This attribute is a standard OpenTelemetry attribute and should be placed on N1QL and analytics operations. The attribute key is "`db.statement`".
+This attribute is a standard OpenTelemetry attribute and should be placed on N1QL and analytics operations. The attribute key is "`db.query.text`". This attribute should only be included if positional or named parameters are in use.
```
brett19, line 599: Why does this omit the query based on there being named parameters?  In the context of spans, it seems like the query text should always be included.

rfc/0067-extended-sdk-observability-rto-v2.md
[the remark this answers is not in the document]
brett19: It's a worthwhile point to make that while OpenTelemetry specifies the units as being seconds, this is merely a transport detail.  In the actual OpenTelemetry tooling, because the units are known, it can be converted visually to something more appealing.

rfc/0067-extended-sdk-observability-rto-v2.md
```diff
@@ -952,12 +970,49 @@ If this information is not available for any reason (generally this will only be
...
+| Peer port | Did not exist | network.peer.port | |
+
+**Notes:**
+
+1. In v0, one of the possible values was "Success". In v1, the attribute should be omitted if there was no error.
+2. In v0, the value was "TCP.IP". In v1, it should be "tcp" instead.
+
+## Opting into the stable semantic conventions
+
+Existing implementations should not change the conventions that they emit by default. Users can opt into the new conventions either using a `observability_stable_semantic_convention_opt_in` (`String[]` type) cluster option. Support should also be added for the `OTEL_SEMCONV_STABILITY_OPT_IN` environment variable (values are comma-separated strings) which is  recognized by many OpenTelemetry instrumentation libraries. If both are set, the cluster option has higher precedence.
```
brett19, line 1001: I think "cluster option" would be less ambiguously represented as `a ClusterOptions field`.  At the moment, on first read this seemed to be implying there would be cluster-side configurability for this.

rfc/0067-extended-sdk-observability-rto-v2.md
```diff
@@ -952,12 +970,49 @@ If this information is not available for any reason (generally this will only be
...
+
+The client should handle switching between the old and the stable conventions internally, and call the relevant methods on the `RequestTracer` and `Meter` interfaces with the right attributes as specified by the option & environment variable above.
+
 # Language Specifics
 
 ## Go
 
-* The Go SDK supports only kv for orphan response logging, all HTTP requests within the Go SDK are performed using the Golang HTTP library, which is synchronous. As such orphaned responses are not possible for non-kv services.
-* The Go SDK architecture is such that on timeouts the requests are removed from internal queueing systems (which is in fact how orphaned responses are detected, the lookup to find the corresponding request fails). This means that only a subset of the orphan logger fields can ever be exposed.
+* The Go SDK supports only KV for orphan response logging, all HTTP requests within the Go SDK are performed using the Golang HTTP library, which is synchronous. As such orphaned responses are not possible for non-kv services.
```
brett19, line 1015: Since HTTP requests are cancellable, unlike KV requests, shouldn't all SDKs have this limitation?  The only way I could imagine getting orphan responses from non-KV services would be if SDKs were continueing to process HTTP requests in the background in spite of them timing out.

rfc/0067-extended-sdk-observability-rto-v2.md
```diff
@@ -604,7 +608,7 @@ This attribute is a standard OpenTelemetry attribute and should be placed on N1Q
 
 #### Attribute: Operation
 
-This attribute is a standard OpenTelemetry attribute and should be placed on all operations which do NOT have the db.statement set. The attribute key is "db.operation".
+This attribute is a standard OpenTelemetry attribute and should be placed on all operations which do NOT have the db.query.text set. The attribute key is "db.operation.name".
```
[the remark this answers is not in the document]
DemetrisChr: I think in the older semantic conventions `db.operation` was specified as conditionally required if `db.statement` is not available, which was probably the motivation for this rule originally. Now, in [v1.37](https://opentelemetry.io/docs/specs/semconv/database/database-spans/#span-definition), they have changed that to 'Conditionally required if readily available', so I think as you say, we should make it universal.

Now the question is what value it should take. From below, we currently have `db.operation.name` defined as follows (just copying it here for convenience):

| Service Type | Value |
| :---- | :---- |
| Data | `<use the span name>` |
| Query | Do not set |
| Search | `search_index_name` |
| Views | `/design_doc_name/view_name` |
| Analytics | Do not set |
| Management | Use the relative path that gets accessed AND the HTTP method (i.e. `"GET /foo/bar/baz"`). So if a scope name of `foo` is queried, it should be `GET /scopes/foo` (as in: include the actual name, not a placeholder like `{scope}`). |

I don't really see something equivalent that would fit for query here, unless we use "query" and "analytics_query".

Somewhat tangentially to this, I'm not sure if our existing values above for `db.operation.name` fit with the spirit of the semantic conventions. They are also different from the `db.operation.name` values we use for the Meter (there `db.operation.name` is the same as the span name for the relevant operation). I feel like two search requests on different search indexes shouldn't have different operation names.

Would it be desirable to have `db.operation.name` be the same as what we currently have as the top-level span names (as we do for metrics)?

Benefits of this are:
* That would remove the need for having different conventions per service
* It would reduce the cardinality of the attribute if there are no bucket names, search index or view names involved.
* If `db.operation.name` is low cardinality, we can use Otel's recommended approach for span names which is `{db.operation.name} {target}`, where target is bucket/scope/collection name.

An issue would be that we lose information on e.g. the search index used, but we could make use [db.operation.parameter.\<key\>](https://opentelemetry.io/docs/specs/semconv/registry/attributes/db/#db-operation-parameter) for that (although it's not yet a stable attribute).

rfc/0067-extended-sdk-observability-rto-v2.md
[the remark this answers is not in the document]
DemetrisChr: > Admittedly on the flip side, it means that on timeout they are looking at a friendlier 2.5 rather than 2500000. But those should be rarer cases.

With this metric we are reporting operation durations across all services, including queries and management operations which are much higher numbers compared to KV.

> metrics are frequently exported to Prometheus which will lose the unit info and will just display 0.000078 (I think)

According to documentation, Prometheus collector translates the Otel unit to the Prometheus equivalent: https://github.com/open-telemetry/opentelemetry-collector-contrib/tree/main/pkg/translator/prometheus#metric-name

dnault (approved):

programmatix (approved):

chvck (approved):

jeffrymorris (approved):

thejcfactor (approved):

avsej (approved):

PR #157, opened 2026-03-20
Adding getIfPresent

programmatix:


rfc/0053-sdk3-crud.md
```diff
@@ -79,6 +79,28 @@ Throws
...
+[IGetResult] GetIfPresent(string id, [options])
+```
+
+It is implemented as a separate API (rather than a toggle in `GetOptions` say) as in the majority of SDKs a nullable/optional return type needs to be part of the type signature.
+
+Parameters: Same as `Get`.
+
+Throws: Same set as `Get`, except `DocumentNotFoundException` which must not be thrown.
+
+Observability: This should be handled exactly as a `Get` operation except named `GetIfPresent`, and spans must not attach `DocumentNotFoundException`.
```
DemetrisChr: NIT: Since we use snake case for operation names in tracing/metrics, should this say that it will be named `get_if_present` instead? It's fairly obvious that this is what it means, but just to avoid any confusion.

dnault (dismissed):

rfc/0053-sdk3-crud.md
```diff
@@ -79,6 +79,28 @@ Throws
 
 - Any exceptions raised by the underlying platform
 
+#### GetIfPresent
```
brett19: This actually sounds like what our current 'Get' should actually be, where we want to get the value if it is present and its otherwise an error, would "GetOrNull" perhaps be less ambiguous?  Additionally, was any consideration made towards implementing this as 'GetOrDefault' instead, which would support users who want it missing to not need to allocate for an exception, but would also provide convenience to allow people to do the same thing with values other than null?

dnault: > **would "GetOrNull" perhaps be less ambiguous**

Data point: `getOrNull` is the idiomatic name for this sort of operation in Kotlin. In fact, the Kotlin SDK already has a `getOrNull`, but it currently just delegates to `get` and catches the exception. It will be nice for it to offer a performance advantage.

I'd be in favor of leaving the exact name up to the SDK implementation in languages where there's a clear precedent.

> **`GetOrDefault' instead, which would support users who want it missing to not need to allocate for an exception, but would also provide convenience to allow people to do the same thing with values other than null?**

In the Java SDK the return type is a `GetResult`, which a user cannot safely create themselves. Being able to specify a default is a cool idea, but I have doubts about whether it's practical in this case.

rfc/0053-sdk3-crud.md
programmatix: Yes good point, will fix.  It is indeed "get_if_present" in the reference impl.

rfc/0053-sdk3-crud.md
programmatix: Yeah I'm not seeing a lot of benefit to GetOrDefault.

I'm open to any reasonable naming.  But baking in "Null" is problematic for languages where it's None or similar instead.

There is David's option of leaving it up to the SDK to find an idiomatic name.

Will leave for a bit and let's see if anyone comes up with a clever name that we can all implement and avoids baking in a specific return type.

programmatix: There is .NET's "tryGet" of course.  It's not perfect because, well, both get methods are going to try and get the doc.  But it beats "getButDoNotThrowAnExceptionButOnlyIfTheDocWasNotFound".

DemetrisChr: `GetIfExists` is another idea (albeit fairly similar to `GetIfPresent`). It might be a bit more consistent with the `Exists` we already have.

Transactions used to have `GetOptional` that had an API similar to this, but I don't know if I'm a massive fan of that name, since not all languages would use an optional type.

programmatix: `unexceptionalGet()`

programmatix: Changed to David's suggestion of getOrNull (with variants allowed), following discussion here https://couchbase.slack.com/archives/G9682CWN7/p1774523856699919

rfc/0053-sdk3-crud.md
```diff
@@ -79,27 +80,32 @@ Throws
 
 - Any exceptions raised by the underlying platform
 
-#### GetIfPresent
+#### GetIfNull
```
dnault: Here and elsewhere, did you mean `GetOrNull` instead of `GetIfNull` ?

dnault (changes requested):

rfc/0053-sdk3-crud.md
programmatix: Ah dang, I did.  Updating.

DemetrisChr (approved):

dnault (approved):

thejcfactor (approved):

PR #159, opened 2026-04-28
Eventing RFC: Add 'Resuming' function status

DemetrisChr:
Eventing has added a `resuming` composite status (equivalent to our `EventingFunctionStatus`) in 8.0. We should also add it to the SDKs. This was first raised as an issue on the Go SDK, see [GOCBC-1791](https://jira.issues.couchbase.com/browse/GOCBC-1791).

[GOCBC-1791]: https://couchbasecloud.atlassian.net/browse/GOCBC-1791?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/rfc/0062-eventing-management.md
+++ b/rfc/0062-eventing-management.md
@@ -891,10 +891,11 @@ enum EventingFunctionStatus {
 	Undeploying  
 	Paused  
 	Pausing  
+	Resuming
 }
 ```
 
-The EventingFunctionStatus mapping from JSON is specified as follows:
+The `EventingFunctionStatus` mapping from JSON is specified as follows:
 
 | State | JSON Property |
 | :---- | :---- |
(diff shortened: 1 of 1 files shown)


rfc/0062-eventing-management.md
```diff
@@ -891,10 +891,11 @@ enum EventingFunctionStatus {
 	Undeploying  
 	Paused  
 	Pausing  
+	Resuming
 }
 ```
 
-The EventingFunctionStatus mapping from JSON is specified as follows:
+The `EventingFunctionStatus` mapping from JSON is specified as follows:
```
programmatix, line 899: Slight scope creep but maybe this is a good time to also add an Unknown status, to provide some level of future compatibility with any other statuses the server adds.

programmatix (dismissed):

rfc/0062-eventing-management.md
DemetrisChr: That's a good point. I'll add it, but with the caveat that not all SDKs will have it. Go (as usual) is a bit different here given its lack of enums. In Go we have the status type as an alias of String with pre-defined constants for the known status values. If an unexpected one is returned, the user can cast it to a string to access the raw status. So we wouldn't need to have an "unknown" status.

DemetrisChr: Added

dnault (approved):

programmatix (approved):

avsej (approved):

thejcfactor (approved):

davidkelly (approved):

jeffrymorris (approved):

PR #160, opened 2026-07-08
RFC-52: Add score fusion (revision 14)

anirudhlakhotia:


rfc/0052-sdk3-full-text-search.md
```diff
@@ -906,6 +906,117 @@ If it does not currently have one (in which case one should already be in the pr
 This may lead the operation to timeout if the config does not arrive in time.
 The SDK will wait for at least the GCCCP config.  To simplify implementations, and because GCCCP has long been available, it may choose to not use bucket configs.
 
+## Score fusion
+This is a feature being added to Couchbase Server 8.1 in the FTS service.  It extends hybrid search, where a traditional FTS query is combined with one or more vector queries.
+
+As with vector search, all SDK additions related to it should be annotated with the platform equivalent of @Stability.Volatile, as changes may be required following user feedback.
```
DemetrisChr: N.B. Vector search is no longer volatile, the section above probably needs to be updated as well.

Do we still want search fusion to be volatile?

rfc/0052-sdk3-full-text-search.md
```diff
@@ -906,6 +906,117 @@ If it does not currently have one (in which case one should already be in the pr
...
+
+References:
+* [Server design document: Hybrid Search Score Fusion](https://docs.google.com/document/d/12Y1txZ6C0gU2v_U9Sicg-7Jsy-ISoZm5VA79rz4kfNw/edit) (MB-60401)
```
programmatix: Nitpick: The MB should be a separate line item.

rfc/0052-sdk3-full-text-search.md
```diff
@@ -906,6 +906,117 @@ If it does not currently have one (in which case one should already be in the pr
...
+* Relative Score Fusion (`"rsf"`).  Merges by normalised score rather than rank.
+
+The score on each returned `SearchRow` is the fused score, not the original FTS or vector score; its magnitude depends on the strategy.
+
+### SearchScoreFusion
+The strategy is selected through a new `SearchScoreFusion` object.  `SearchScoreFusion` creation is platform-idiomatic:
+
+```
+SearchScoreFusion.rrf()
+SearchScoreFusion.rsf()
```
DemetrisChr: This is subjective, but I find these names too cryptic. What do you think about the following?
```
SearchScoreFusion.reciprocalRank()
SearchScoreFusion.relativeScore()
```

rfc/0052-sdk3-full-text-search.md
```diff
@@ -906,6 +906,117 @@ If it does not currently have one (in which case one should already be in the pr
...
+```
+
+Each strategy's tuning parameters are exposed in the same way the SDK exposes traditional FTS query parameters (fluent-style methods on the object, or an options block, following the SDK's existing convention):
+
+* `rankConstant` (`uint32`, RRF only).  Sent as `params.score_rank_constant`.  The server defaults it to `60`.  Not exposed on `rsf()`.
+* `windowSize` (`uint32`).  Sent as `params.score_window_size`.  How many results per list are considered for fusion.  The server defaults it to the request `size` (`limit`).
```
programmatix: It does?  Weird.  Seems the whole point is to have a significantly larger window than the final result size, so you know you're fusing a good set of results from both sources.

rfc/0052-sdk3-full-text-search.md
```diff
@@ -906,6 +906,117 @@ If it does not currently have one (in which case one should already be in the pr
...
+
+`rrf()` exposes both `rankConstant` and `windowSize`; `rsf()` exposes only `windowSize`.  As with `numCandidates` (`k`) and `knn_operator`, the SDK names differ from the JSON field names.  In SDKs where a single class exposes both `rrf()` and `rsf()` rather than separate types, `rankConstant` cannot be hidden on `rsf()`; if it is set there, the SDK must ignore it and not send it.
```
programmatix: > where a single class exposes both `rrf()` and `rsf()`

I would explicitly ban that here.  It already doesn't really work, and won't scale if more scoring options get added later.

rfc/0052-sdk3-full-text-search.md
```diff
@@ -906,6 +906,117 @@ If it does not currently have one (in which case one should already be in the pr
...
+### Errors and failure states
+The SDK keeps client-side validation minimal: before sending, it checks only that `disableScoring(true)` and `scoreFusion` are not both set.  The server validates everything else, including parameter ranges and disallowed sorts.  The failure states, with the error each produces:
+
+* `disableScoring(true)` and `scoreFusion(...)` both set.  The SDK raises `InvalidArgumentException` before sending; this is the only combination it validates.
+
+* `sort` other than the default `"-_score"` under fusion.  The SDK does not check this; the server returns a 400 and a `CouchbaseException` is raised.
+
+* Cluster does not support fusion.  Detectable only through the cluster capability (see FeatureNotAvailable handling).  Once that capability exists, the SDK raises `FeatureNotAvailableException` before sending; until then, the request silently degrades to additive scoring with no error.
+
+* Non-hybrid request (only an FTS query, or only a vector search).  The SDK does not check for this; it sends the `score` field regardless, and the request is a silent no-op rather than an error.
```
programmatix: While not a blocker, personally I would InvalidArgument this.  It just helps guide the user to the correct use of a complex FTS API.

rfc/0052-sdk3-full-text-search.md
```diff
@@ -906,6 +906,117 @@ If it does not currently have one (in which case one should already be in the pr
...
+
+### couchbase2
+Score fusion only affects a hybrid request, and couchbase2 does not implement vector search (see the vector search section above).  A hybrid request is therefore rejected there for its vector search, and `scoreFusion` on any other request is the same no-op as elsewhere.  No additional couchbase2 handling is required.
+
+### FeatureNotAvailable handling
+There is no cluster capability for score fusion yet, so the first release does not gate it: a `scoreFusion` request is sent to any cluster.  A pre-fusion server ignores any `score` value other than `"none"`, so the request returns 200 ranked with normal additive scoring, with no error.  Nothing in the response tells the SDK or the user that fusion was ignored.
+
+When ns_server advertises a capability for score fusion under `clusterCapabilities.search`, the SDK will gate on it: whenever `scoreFusion` is set, it checks the capability before sending and raises `FeatureNotAvailableException` if the cluster does not advertise it, the same way the vector-search capability is checked.
```
DemetrisChr: Are there plans for ns_server to introduce this capability?

Considering that SDKs are not expected to implement this yet, I think it makes sense to include this in the RFC only when ns_server implements it and SDKs are expected to add the check.

rfc/0052-sdk3-full-text-search.md
```diff
@@ -906,6 +906,117 @@ If it does not currently have one (in which case one should already be in the pr
...
+
+* Non-hybrid request (only an FTS query, or only a vector search).  The SDK does not check for this; it sends the `score` field regardless, and the request is a silent no-op rather than an error.
+
+* Out-of-range `rankConstant` or `windowSize`.  The SDK does not validate parameter ranges; it forwards the value, the server returns a 400, and a `CouchbaseException` is raised.
+
+For transport, a score-fusion request is an ordinary search: timeouts and service failures surface as `TimeoutException` or `CouchbaseException`, exactly as they do for `cluster/scope.search()`.
+
+### couchbase2
+Score fusion only affects a hybrid request, and couchbase2 does not implement vector search (see the vector search section above).  A hybrid request is therefore rejected there for its vector search, and `scoreFusion` on any other request is the same no-op as elsewhere.  No additional couchbase2 handling is required.
```
DemetrisChr: I think couchbase2 now supports vector search. See: https://github.com/couchbase/protostellar/blob/master/couchbase/search/v1/search.proto#L440-L444

anirudhlakhotia: Ah, missed this, thanks! 
Yep, we'll raise `FeatureNotAvailableException` if scoreFusion is set on a couchbase2 request

rfc/0052-sdk3-full-text-search.md
anirudhlakhotia: > Are there plans for ns_server to introduce this capability?

No firm commitment from ns_server yet, I'm planning to raise it with them soon. 
 
 > I think it makes sense to include this in the RFC only when ns_server implements it and SDKs are expected to add the check.
 
Makes sense. Just something to keep in mind, since until the cap exists it silently falls back to additive scoring rather than erroring.

rfc/0052-sdk3-full-text-search.md
anirudhlakhotia: Hmmm, thought about this for a bit. None of the SDKs abbreviate feature names in the API surface (it's always spelled out - matchPhrase, numericRange, geoDistance, never the short form), so rrf/rsf would stick out.

I also checked what the others do - elastic and qdrant both call it rrf, weaviate spells it out. But nobody uses "rsf" - qdrant's second algorithm is a different one (dbsf), and weaviate spells it relativeScoreFusion. So rrf is a real, recognised term and rsf isnt.
 
So overall I'm leaning towards `reciprocalRank() / relativeScore()` too.  Open to other opinions though.

rfc/0052-sdk3-full-text-search.md
DemetrisChr: Based on the discussion below for the ns_server cap, I think it would make sense to keep search fusion volatile at least until the ns_server cluster cap is added. When we implement the cluster cap check it will technically be a breaking change, so if it's 'volatile' that would be acceptable.

rfc/0052-sdk3-full-text-search.md
```diff
@@ -906,6 +906,117 @@ If it does not currently have one (in which case one should already be in the pr
...
+
+cluster.search("products_index", request,
+        SearchOptions.searchOptions().scoreFusion(SearchScoreFusion.rrf()));
+```
+
+`scoreFusion` and `disableScoring(true)` both write the top-level `score` field, so they cannot be set together.
+
+If neither is set, the SDK sends no `score` field and the server keeps its existing additive-boost behaviour, so existing queries are unaffected.
+
+A unified `scoring(...)` setter was rejected: `disableScoring` is already shipped across the SDKs, so unifying would force a deprecation everywhere.
```
dnault: A unified `scoring(...)` setter sounds great to me (at least in Java and Kotlin and any other language where it makes sense). Is there a link to a discussion somewhere that might change my mind?

rfc/0052-sdk3-full-text-search.md
```diff
@@ -906,6 +906,117 @@ If it does not currently have one (in which case one should already be in the pr
...
+
+### The boost field as a fusion weight
```
dnault, line 985: Does an SDK developer need to know the info in this section to implement the feature, or is it just background information?

rfc/0052-sdk3-full-text-search.md
anirudhlakhotia: Yup this makes sense. The cap check landing later is a breaking change (silent fallback becomes an exception), and volatile lets us add it without a deprecation cycle

anirudhlakhotia: Filed [MB-72837](https://jira.issues.couchbase.com/browse/MB-72837) for this

rfc/0052-sdk3-full-text-search.md
programmatix: Yeah agreed, the verbose option is better

rfc/0052-sdk3-full-text-search.md
programmatix: OOI what would you put in there?

dnault: > **OOI what would you put in there?**

@programmatix When I wrote that comment, I was thinking it would follow the usual Scala/Kotlin pattern:

```
scoring(SearchScoring.none()  / SearchScoring.somethingElse())
```

but I'm backing away from this, since it doesn't really solve the problem of invalid arguments.

For Kotlin, we can make it impossible to specify an invalid arg; the scoring fusion can be an optional parameter of the method that builds the hybrid query. For example:

```
cluster.search(
    indexName = "myIndex",
    spec = SearchSpec.mixedMode(
        searchQuery = SearchSpec.queryString("foo"),
        vectorQuery = SearchSpec.vector("someField", vector),
        scoreFusion = SearchScoreFusion.rsf(windowSize = ...)   // <-- HERE!
    )
)
```

rfc/0052-sdk3-full-text-search.md
anirudhlakhotia: Still unsure about `Volatile` to be honest, mainly because of the "boost" reuse
But otherwise resolved with the new ns_server cluster cap

What does everyone think about this?

rfc/0052-sdk3-full-text-search.md
programmatix: Nb I like a unified scoring method, in discussions with @anirudhlakhotia and think this is going to make it in.
Something like `SearchOptions.searchOptions().scoring(SearchScoring.reciprocalRank())`

anirudhlakhotia: We would need to soft deprecate `disableScoring` across all SDKs though, but this is clean and lets us extend the method to other non-fusion scoring methods if we have any in the future

DemetrisChr: I was looking at [CBD-3575](https://jira.issues.couchbase.com/browse/CBD-3575) which is when `disableScoring` was introduced. It says:
> If in the future we have a need to add the BM25 option or other enumerations on the option block, we would deprecate the disableScoring() and add that enumeration to the options.   The "raw" for other stuff remains a workaround if there is another surprise change from the FTS service.

This could mean that we might eventually have other scoring functions, that aren't necessarily a form of score fusion (do we know if such functions already exist?), and could also be set for standard search queries without vector search.

So this might not just be just for including `disableScoring`, but it would leave the door open to non-fusion scoring functions, without adding another option.

Though if we do this, we should change the naming slightly to `SearchScoring.reciprocalRankFusion()` and `SearchScoring.relativeScoreFusion()`, as "Fusion" would not be part of the class name.

Matt-Woz: +1 for a unified scoring method. Also,`disableScoring` was not codified in the RFC before this change

anirudhlakhotia: Perfect
I was just going to update the RFC with the unified scoring method too

> (do we know if such functions already exist?), and could also be set for standard search queries without vector search.

Custom scoring is on the FTS backlog as [IDEA-794](https://jira.issues.couchbase.com/jira/polaris/projects/IDEA/ideas/view/6220656?selectedIssue=IDEA-794), and [MB-65018](https://jira.issues.couchbase.com/browse/MB-65018) discusses a script_score equivalent using JS UDFs

> Though if we do this, we should change the naming slightly to SearchScoring.reciprocalRankFusion() and SearchScoring.relativeScoreFusion(), as "Fusion" would not be part of the class name.

Yeah I like this style of naming more with the member methods showing that they're fusion types.

rfc/0052-sdk3-full-text-search.md
DemetrisChr: I think there's no need for this to be volatile, now that we have the cap.

Though the convention is that we keep it 'uncommitted' until the SDK dot-minor that aligns with the server release, but that's not a detail that the RFC needs to mention, so I think removing the section that talks about stability should be fine.

programmatix (approved):

dnault (approved):

rfc/0052-sdk3-full-text-search.md
```diff
@@ -906,6 +906,129 @@ If it does not currently have one (in which case one should already be in the pr
...
+### SearchScoring
+The scoring mode is selected through a new `SearchScoring` object. `SearchScoring` creation is platform-idiomatic:
+
+```
+SearchScoring.reciprocalRankFusion()
+SearchScoring.relativeScoreFusion()
+SearchScoring.none()
+```
+
+`reciprocalRankFusion()` and `relativeScoreFusion()` are the two fusion strategies; the names keep "fusion" since `SearchScoring` itself is not fusion-specific. `none()` disables scoring, sending the same `"none"` that `disableScoring(true)` sends today; it is not a fusion strategy and works on any server version.
```
DemetrisChr: > the names keep "fusion" since `SearchScoring` itself is not fusion-specific.

NIT: This doesn't make sense outside the context of the comment thread, since there's no suggestion of a name without "fusion" in the final version of the RFC. I think the RFC doesn't always need to justify the decisions to this extent.

DemetrisChr (approved):

rfc/0052-sdk3-full-text-search.md
anirudhlakhotia: Mostly background - there isn't anything new for the SDKs to implement here, since weights deliberately reuse the existing `boost` API. I kept the section because it answers "where do weights go?", and because the
meaning of the top-level boost changes under fusion. I think the SDK docs might need to explain that distinction to users

PR #161, opened 2026-07-16
Adding GetReplica API

programmatix:


rfc/0053-sdk3-crud.md
```diff
@@ -629,6 +630,84 @@ The SDK can then detect the marker interface, fetch the string value, set the Xa
...
+
+###### `GetReplicaStrategy.fromIndex`
+The initially supported strategy is:
+
+```
+GetReplicaStrategy.fromIndex(ReplicaIndex.FIRST, [GetReplicaStrategyFromIndexOptions options])
+```
+
+Where `ReplicaIndex` is an enum of `FIRST`, `SECOND`, `THIRD`.  
+Not supported until there is a concrete use-case for it is `ACTIVE`:  the user should use regular `Get`, and can use `GetReplica` on exceptions.
```
davidkelly, line 694: I recall a discussion of a RANDOM index as well - which one could do by knowing how many replicas currently and deciding what to put in here, but I wonder if that is needed?   Not pushing either way, just curious.

rfc/0053-sdk3-crud.md
```diff
@@ -26,6 +26,7 @@ interface ICollection{
 
   IGetResult Get(string id, GetOptions options = null);
   [IGetResult] GetOrNull(string id, GetOptions options = null);
+  [IGetReplicaResult] GetReplica(string id, GetReplicaStrategy strategy, GetReplicaOptions options = null);
```
DemetrisChr: Will this ever return `null`? The `[]` surrounding `IGetReplicaResult` suggests that, but it doesn't seem to be the case, since this throws `DocumentNotFoundOnReplicaException`.

rfc/0053-sdk3-crud.md
```diff
@@ -629,6 +630,84 @@ The SDK can then detect the marker interface, fetch the string value, set the Xa
...
+  - Id: string - the primary key.
+
+- Optional
+  
+  - Timeout - the time allowed for the operation to be terminated.  Defaults to the same as regular Gets.
+  - Transcoder - a custom transcoder.
+
+- Throws
+  - Documented
+    - DocumentNotFoundOnReplicaException
```
DemetrisChr, line 658: Is a new exception type necessary for this? This will never return `DocumentNotFoundException`. Could we not use that instead? I would imagine that users would have a narrow try-catch block around the `GetReplica` operation if they want to do something in response to that, so it wouldn't get mixed up with a `DocumentNotFoundException` thrown by the standard `Get`.

Separate exceptions would also be a departure from the current approach where the mapping of KV status to exception type is shared between all operations. Both `DocumentNotFoundException` and `DocumentNotFoundOnReplicaException` correspond to the 0x01 (not found) KV status.

I don't feel strongly about this, just curious about your thoughts.

rfc/0053-sdk3-crud.md
```diff
@@ -629,6 +630,84 @@ The SDK can then detect the marker interface, fetch the string value, set the Xa
...
+      - A new exception for this feature.  It should extend/inherit from DocumentNotFoundException.
+    - RequestTimeoutException
```
DemetrisChr: We don't have an exception type called `RequestTimeoutException`

rfc/0053-sdk3-crud.md
```diff
@@ -629,6 +630,84 @@ The SDK can then detect the marker interface, fetch the string value, set the Xa
...
+```
+GetReplicaStrategy.fromIndex(ReplicaIndex.FIRST, [GetReplicaStrategyFromIndexOptions options])
+```
+
+Where `ReplicaIndex` is an enum of `FIRST`, `SECOND`, `THIRD`.  
+Not supported until there is a concrete use-case for it is `ACTIVE`:  the user should use regular `Get`, and can use `GetReplica` on exceptions.
+
+GetReplicaStrategyFromIndexOptions options:
+
+- `Boolean wrap` - whether to wrap around the available replicas, rather than throwing `ReplicaIndexOutOfBoundsException`.  Defaults to false.
```
dnault, line 698: Why would a user want to wrap instead of being told the replica doesn't exist?

rfc/0053-sdk3-crud.md
```diff
@@ -629,6 +630,84 @@ The SDK can then detect the marker interface, fetch the string value, set the Xa
...
+
+- Optional
+  
+  - Timeout - the time allowed for the operation to be terminated.  Defaults to the same as regular Gets.
+  - Transcoder - a custom transcoder.
+
+- Throws
+  - Documented
+    - DocumentNotFoundOnReplicaException
+      - A new exception for this feature.  It should extend/inherit from DocumentNotFoundException.
```
avsej, line 659: could it be just DocumentNotFoundException? why do we need this specific exception?

rfc/0053-sdk3-crud.md
```diff
@@ -629,6 +630,84 @@ The SDK can then detect the marker interface, fetch the string value, set the Xa
...
+`GetReplica` is a forward-looking API designed to:
+
+a) provide the user with a fundamental building block in `GetReplicaStrategy.fromIndex` upon which they can build any strategy they choose.
+b) a plugin-based flexible approach that allows Couchbase to add new single-replica strategies in the future without adding another Collection-level API.
+c) solve the issues of the existing `GetAnyReplica` method.
+
+[Design notes](https://github.com/couchbaselabs/sdk-design/blob/main/2026/get-replica) (private to Couchbase employees) are available for optional further context.
+
+Strategies can read multiple replicas, and the active, though of course per the API must ultimately return a single `GetReplicaResult`.   
+Future strategies could include one that replaces `GetAnyReplica` (at which point that method will be deprecated) such as `GetReplicaStrategy.fastest()`.
```
avsej, line 682: Is there alternative name to `fastest` that does not carry "performance guarantee" flavour? How about `GetReplicaStrategy.returnFirst()`?

rfc/0053-sdk3-crud.md
avsej: Cluster deployment has only two replica, but application request `THIRD`, wrap here would be something like `3 % 2 => 1`. I think if we have wrap, then we need `clamp` too, which will give me something like `max(bucket.num_replicas, replica)`, here `max(2, 3) => 2`.

also, how about case where the number of replicas satisfies the requested index, but the index resolves into node index `-1`? is this the same? should we wrap/clamp?

rfc/0053-sdk3-crud.md
programmatix: Yes there is scope for that though I would lean to implementing it as `RetryStrategy.random()`.

On operational it's not strictly required, the user can easily loop through `fromIndex` until fast-failure.  On couchbase2:// it does have utility as it can avoid round-trips to find the last replica (assuming we end up being able to push it down in the GRPC).  But the `wrap` fromIndex parameter also handles the same case.  So it would be just sugar - but it's nice sugar.

rfc/0053-sdk3-crud.md
programmatix: Oops no, that's a C&P error.  Will fix.

rfc/0053-sdk3-crud.md
programmatix: Yep fair question.  See what you think of my justifications:

> This will never return DocumentNotFoundException

It can, as future strategies are allowed to target the active.  (Thinking of use-cases like preferred server groups here)
In which case it's very useful for the user to know if the signal came from the active or not - analogous to the existing isReplica on GetReplicaResult.  `DocumentNotFoundException` is a much stronger signal than `DocumentNotFoundOnReplicaException` and could deserve separate programmatic treatment (the primary SDK3 criteria for whether a new exception is warranted).  Though tbh I don't have a concrete use-case for that separate treatment in mind currently.

I also like that it reminds users that replicas are eventually consistent.

rfc/0053-sdk3-crud.md
programmatix: Ah, this was copied from the existing GetAnyReplicas block and I assumed that was accurate as it's been in there for years.  Yes, will fix.

rfc/0053-sdk3-crud.md
programmatix: @dnault primarily to support the random use-case, and primary on couchbase2://.  Avoids an 'error hop', and avoids the user needing to code in knowledge of their topology setup into the app.

@avsej I'm not sure about a `clamp` as it would bias things towards the last node, in the random use-case.  Do you have a use-case in mind for it?

-1 was discussed on Slack including w.r.t. wrap - have tightened up the RFC on it; let me know what you think.

rfc/0053-sdk3-crud.md
programmatix: Asked/discussed in another conversation, so will resolve this one.

rfc/0053-sdk3-crud.md
programmatix: Nb this strategy isn't currently in this design, it's just an example.

But yes your point is fair.  Users might be using this to maximise chance of getting a result, not just a fast one.
Other terms I've seen for this are: hedging; racing; speculative execution.

We can defer all that until the strategy is designed though.

rfc/0053-sdk3-crud.md
DemetrisChr: This sounds reasonable, especially considering that eventually we might also be fetching the active from here. I think I'm convinced

rfc/0053-sdk3-crud.md
```diff
@@ -629,6 +630,88 @@ The SDK can then detect the marker interface, fetch the string value, set the Xa
...
+
+- Optional
+  
+  - Timeout - the time allowed for the operation to be terminated.  Defaults to the same as regular Gets.
+  - Transcoder - a custom transcoder.
+
+- Throws
+  - Documented
+    - DocumentNotFoundOnReplicaException
+      - A new exception for this feature.  It should extend/inherit from DocumentNotFoundException.
```
DemetrisChr, line 659: In C++, we might not have something equivalent to the inheritance required here, because we just use error codes to distinguish between the various error types.

The RFC probably can't do anything about this, but I thought I'd just highlight this.

programmatix: Indeed, this isn't something I can accommodate, or tbh should have to.  C++ has decided to deviate massively from the intended SDK error model (0058-error-handling.md) and will have to find its own path here.

DemetrisChr (approved):

emilienbev (approved):

avsej (approved):

PR #162, opened 2026-07-23
Revision 21: Handling query error code 2120

programmatix:


dnault (approved):
Unfortunate that we're once again conflating authentication and authorization, but them's the breaks.

Matt-Woz (approved):

emilienbev (approved):

PR #163, opened 2026-09-10
RFC-0086: Conversational Query

anirudhlakhotia:
Adds RFC 86, Conversational Query, as `rfc/0086-conversational-query.md` (revision 3), plus a row in the README draft index.

Owner is Jared Casey; revision 3 restructures the earlier draft to follow the SDK RFC layout and consolidates the public API, model credentials/options, chat lifecycle, `ChatHandle`, Knowledge management, wire mapping, errors, retries, and feature detection.

Opened as a draft for review of the API shape before signoff. Note the README still says "Next RFC ID 85" — left alone in case 85 is claimed elsewhere.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/rfc/0086-conversational-query.md
+++ b/rfc/0086-conversational-query.md
@@ -1,0 +1,714 @@
+# Meta
+
+* RFC Name: Conversational Query
+* RFC ID: 86
+* Start Date: 2026-08-31
+* Owner: Jared Casey
+* Current Status: DRAFT
+* Current Editor: Anirudh Lakhotia
+* Revision: 3
+* Supporting Material: [sdk-design](https://github.com/couchbaselabs/sdk-design/tree/main/server-aligned/totoro/conversational-query)
+
+# Summary
+
+Conversational Query allows applications to submit natural-language prompts to the Query service and
+receive generated SQL++ statements and Query results. This RFC defines one-shot queries, multi-turn
+chats, model credentials, Knowledge management, and the SDK behavior required to continue or persist
+a chat across application requests.
+
+# Motivation
+
+Applications may need to continue a conversation across requests handled by different processes. A
+live chat resides on one Query node, so the SDK must retain enough client state to address it from a
+new SDK instance.
+
+# General Design
+
+Conversational Query uses the Query service to generate SQL++ from a natural-language prompt and,
+when requested, execute the generated statement under the caller's Couchbase identity. Model access
+uses separate model credentials.
+
+A live chat is held in memory by one Query node. The SDK retains its logical owner for routing and can serialize
+the required client state in a `ChatHandle`. `chat(handle)` reconstructs a `Chat` from a previously
+serialized handle and performs no Query request. Persisting a chat requires PAUSE; RESUME restores a
+paused chat to live server state.
+
+Chat discovery, model-provider discovery, typed EXPLAIN/ADVISE actions, a public chat-id accessor,
+single-entry Knowledge lookup, and changes to the inactivity timeout after BEGIN are outside this
+revision.
+
+## Public API
+
+```
+Keyspaces = list<string>  // SQL++ keyspace paths
+
+Cluster.conversationalQuery([credential ModelCredential], [model ModelOptions])
+    -> ConversationalQuery
+
+ConversationalQuery
+    query(prompt string,
+          keyspaces Keyspaces,
+          [options QueryOptions],
+          [credential ModelCredential])
+        -> ConversationalQueryResult
+
+    beginChat(keyspaces Keyspaces,
+              [options BeginChatOptions],
+              [credential ModelCredential])
+        -> Chat
+
+    chat(handle ChatHandle,
+         [credential ModelCredential])
+        -> Chat                       // local operation; performs no server I/O
+
+    knowledge() -> KnowledgeManager
+
+KnowledgeManager
+    upsert(name string, keyspace string, value string)   // CREATE OR REPLACE
+    create(name string, keyspace string, value string)   // CREATE; fails if it exists
+    drop(name string, keyspace string)
+    getAll([keyspace string]) -> KnowledgeEntry[]
+
+KnowledgeEntry {
+    name string
+    keyspace string
+    value string
+}
+
+Chat
+    ask(prompt string, [options AskOptions])   -> ConversationalQueryResult
+    pause([options PauseChatOptions])          -> ConversationalQueryMetaData
+    resume()                                   -> ConversationalQueryMetaData
+    end()                                      -> ConversationalQueryMetaData
+
+    handle() -> ChatHandle
+    state()  -> ChatState
+
+ModelCredential
+    capella(email string, password string, organizationId string)
+    stored(credentialName string)
+    apiKey(key string)
+
+ModelOptions {
+    provider string
+    name string
+    endpoint string
+    region string
+    outputTokenLimit int
+    moderation optional<bool>
+    raw map<string, JsonValue>
+}
+
+QueryOptions {
+    model ModelOptions
+    hint string
+    execute bool
+    output ConversationalQueryOutput
+    knowledge bool
+}
+
+BeginChatOptions {
+    inactivityTimeout Duration
+    knowledge bool
+}
+
+AskOptions {
+    hint string
+    execute bool
+    output ConversationalQueryOutput
+}
+
+PauseChatOptions {
+    summarize optional<bool>
+}
+
+ConversationalQueryOutput {
+    Sqlpp
+    FtsSqlpp
+    JsUdf
+}
+
+ChatState {
+    Active
+    Paused
+    Indeterminate
+}
+```
+
+Common Query request controls and asynchronous forms follow existing SDK conventions.
+
+## Model Credentials
+
+`ModelCredential` authenticates Query's calls to the model provider. The SDK must not use it for
+Couchbase authentication or accept it in `ClusterOptions`.
+
+| Credential | Request parameters |
+| --- | --- |
+| `capella(email, password, organizationId)` | `natural_cred` = `email:password`; `natural_orgid` = organization id |
+| `stored(credentialName)` | `natural_config.cred_id` |
+| `apiKey(key)` | `natural_config.api_key` |
+
+A credential supplied to `conversationalQuery` is the default. An explicit credential on `query`,
+`beginChat`, or `chat(handle)` replaces that default for the call or returned `Chat`.
+
+Query does not store model credentials with a live or paused chat. The SDK therefore sends the
+effective credential on each request that may invoke the model. BEGIN, RESUME and END do not invoke
+the model and do not carry model credentials.
+
+The three credential forms are mutually exclusive. The SDK must not:
+
+* combine Capella credentials with `stored` or `apiKey`;
+* send both `cred_id` and `api_key`; or
+* fall back to another credential form after a failure.
+
+These rules apply to the assembled request because Query silently applies credential precedence;
+Capella credentials cause it to ignore `natural_config`.
+
+If no model credential remains after applying defaults and overrides, Query uses the direct-provider
+path and validates whether the resulting configuration can authenticate. The SDK must not reproduce
+provider-specific authentication rules or introduce additional credential kinds.
+
+For `capella`, Query performs the Capella authentication and completion calls. The SDK does not call
+Capella directly.
+
+SDKs must not inspect or change cluster configuration to pre-validate model-credential prerequisites.
+Those prerequisites are documented under [Documentation](#documentation); failures use normal Query
+error handling.
+
+Secrets must be sent as request parameters, never embedded in statement text. The SDK must redact
+`natural_cred`, `natural_config.api_key`, and Capella passwords from errors, error contexts,
+diagnostics, logs, and tracing.
+
+`stored` sends only the name of an existing server credential as `cred_id`, not its secret. The SDK
+does not create or update stored credentials.
+
+## Model Options
+
+All `ModelOptions` fields are optional.
+
+`provider` and `name` are open string values. SDKs must not reject them based on a hard-coded list of supported providers or models.
+
+* `provider`: provider identifier. If omitted, the direct-provider path may use a server default.
+  Query rejects unsupported provider names.
+* `name`: model identifier. Query passes it verbatim to the selected provider without validating
+  that the model exists.
+* `endpoint`: complete completions URL for the direct-provider path. Query appends no suffix. An
+  explicit endpoint currently permits configurations without a model credential.
+* `region`: provider region; currently used by the Bedrock path.
+* `outputTokenLimit`: output-token limit for the direct-provider path.
+* `moderation`: optional boolean. Unset omits the field and uses Query's default moderation behavior.
+  Explicit `true` or `false` preserves the caller's choice.
+* `raw`: additional top-level `natural_config` entries; values may be any JSON value.
+
+A `ModelOptions` supplied to `conversationalQuery` is the default model configuration. A `Chat`
+created by `beginChat` uses those defaults. A `Chat` reconstructed with `chat(handle)` uses the model
+defaults of the attaching `ConversationalQuery`; the handle contains no model configuration.
+
+For a one-shot query, `QueryOptions.model` replaces that default as a whole, including `raw`;
+fields are not merged. An explicitly empty replacement inherits no model fields or raw entries.
+
+Credential selection is independent of model replacement.
+
+### Raw model options
+
+`raw` exists only as a forward-compatibility mechanism for future Query fields. Current Query
+versions silently discard unrecognized `natural_config` keys; `raw` is not provider passthrough.
+
+The SDK must reject these keys in `raw` because they already have typed representations:
+
+`provider`, `model`, `endpoint`, `region`, `output_token_limit`, `moderation`, `cred_id`, `api_key`.
+
+This applies even when the corresponding typed option or credential is unset. Allowing the same
+setting through two paths would make precedence SDK-specific.
+
+Other `raw` entries are added unchanged to `natural_config`. They cannot override Query-controlled
+settings such as temperature or seed.
+
+The SDK must reject options with no wire representation for the selected credential path; see
+[Wire Mapping](#wire-mapping).
+
+## One-shot Query
+
+`query` submits a single prompt with at least one caller-supplied keyspace that Query may use. It
+returns a `ConversationalQueryResult` containing the generated SQL++ statement, when supplied, and
+any execution results. Each call is independent, uses no chat state, creates no `ChatHandle`, and
+requires no Query-node affinity.
+
+Execution is enabled by default. `execute(false)` requests SQL++ generation without execution.
+
+## Chat Lifecycle
+
+A live chat is held in memory by one Query node. It can disappear if that node restarts, is failed over or rebalanced out, or if the chat is evicted or expires from inactivity. Unless the chat has been paused first, that state is not persisted elsewhere.
+
+A paused chat is stored persistently by Query and has its own server-controlled expiry. SDKs must not try to predict when live or paused chats expire or reject requests based on hard-coded server limits.
+
+Applications must not perform multiple operations on the same chat concurrently. SDKs are not required to coordinate concurrent use of separate `Chat` instances or copies of the same `ChatHandle`.
+
+### State
+
+`Chat.state()` reports the state the SDK last observed for the chat. It does not query the server or guarantee that the chat is still in that state.
+
+| State | Meaning |
+| --- | --- |
+| `Active` | The SDK last saw the chat as live and knows which Query node owns it. |
+| `Paused` | PAUSE succeeded, so the SDK no longer treats any Query node as the live owner. |
+| `Indeterminate` | The SDK sent PAUSE but lost the response, so it does not know whether the chat is still live or was successfully paused. |
+
+`Indeterminate` prevents another `ask` because PAUSE may already have removed the live chat from its owner.
+
+| Operation | Permitted local state | Effect of a successful response |
+| --- | --- | --- |
+| `ask` | Active | Stay Active. |
+| `pause` | Active | Become Paused. |
+| `resume` | Paused or Indeterminate | Become Active; the Query node that serves RESUME becomes the new owner. |
+| `end` | Active or Indeterminate | End the chat. |
+
+Operations outside their permitted local states fail with the SDK's normal invalid-operation error.
+
+### Begin
+
+`beginChat` creates a new live chat. The SDK sends BEGIN to any Query node with the requested keyspaces and optional inactivity timeout. The node that serves the request becomes the chat's owner.
+
+Unlike one-shot `query`, the SDK does not require the keyspace list to be nonempty. Query validates the supplied keyspaces and inactivity timeout.
+
+BEGIN does not call the model. Model options, model credentials, and the Knowledge choice are therefore not sent with the BEGIN request.
+
+The SDK keeps the chat's Knowledge choice for later `ask` calls. A credential passed to `beginChat` similarly becomes the credential used by the returned `Chat` when a later operation needs the model.
+
+### Ask
+
+`ask` sends a prompt to an Active chat. The SDK must send it to the Query node that owns the chat.
+
+The chat's keyspaces were set by BEGIN and cannot be changed for an individual turn. The SDK sends the chat's Knowledge choice with every `ask`. `execute` and `output` behave the same way as for a one-shot query.
+
+A hint applies only to the current turn and is not retained as chat configuration. Its text remains
+in conversation history until that history is summarized or rebuilt.
+
+A failed `ask` leaves the local state Active, but may already have added the prompt to the
+conversation. The [non-replay rule](#retries) applies.
+
+### Pause
+
+`pause` saves the conversation so it can later be resumed and releases the chat's live slot on its Query node.
+
+After a successful PAUSE, the SDK marks the chat Paused and no longer keeps a live owner for it.
+
+`summarize` controls whether Query may summarize the conversation before saving it:
+
+- unset: let Query decide;
+- `true`: request summarization;
+- `false`: do not summarize.
+
+If summarization is possible, PAUSE may invoke the model, so the SDK sends the chat's model configuration and available model credential. When `summarize` is `false`, it sends neither. Because PAUSE may include a model call, its default request timeout must not be shorter than the SDK's model-turn timeout.
+
+If Query returns a PAUSE error, the SDK keeps the chat Active and retains its owner. This does not guarantee that the conversation history is unchanged or that no persisted copy exists: summarization may already have replaced the live history before persistence failed.
+
+If a PAUSE request may have reached Query before a transport failure, the outcome is handled under [Ambiguous Outcomes](#ambiguous-outcomes).
+
+### Resume
+
+`resume` turns a Paused chat back into a live chat.
+
+For a Paused chat, the SDK may send RESUME to any Query node. If it succeeds, that node becomes the new owner and the SDK marks the chat Active.
+
+If Query returns an error, the SDK keeps the chat Paused. Neither error 19243 nor metadata error 4500 proves that the persisted chat no longer exists.
+
+For an Indeterminate chat, RESUME instead targets the Query node retained from the ambiguous PAUSE. The additional rules, including the 19255 case, are described under [Ambiguous Outcomes](#ambiguous-outcomes).
+
+The SDK must not use RESUME as a way to check whether a chat is paused. A successful RESUME changes server state by consuming the persisted chat and creating a new live chat.
+
+### End
+
+`end` permanently ends a live chat and releases its live slot. The SDK sends END to the Query node that owns the chat.
+
+END does not delete a paused chat. The SDK must reject `end()` in Paused state and must not
+implicitly resume it. The caller may explicitly resume and end the chat, or let it expire.
+
+After END succeeds, the `Chat` can no longer be used. The SDK must reject further server operations, and `handle()` and `state()` are no longer valid for that object.
+
+Handles that were serialized earlier cannot be revoked. Applications should therefore discard saved handles after successfully ending a chat.
+
+### Ambiguous Outcomes
+
+If the SDK sends PAUSE but loses the response after the request may have reached Query, it cannot know whether the chat is still live or was successfully paused. The SDK therefore marks the chat `Indeterminate`.
+
+The handle keeps the chat id, Knowledge choice, and the Query node that owned the chat before PAUSE. That node becomes the target for resolving the ambiguous PAUSE. If the SDK knows the request failed before it could reach Query, the chat remains `Active` instead.
+
+From `Indeterminate`, the caller may explicitly call `resume()` or `end()`. Both operations must target the Query node retained from the failed PAUSE. The SDK must not search other Query nodes, because another node cannot establish what happened to the PAUSE request on the original owner.
+
+If that Query node can no longer be resolved from the current topology, the SDK returns a routing error and keeps the chat `Indeterminate`.
+
+A successful RESUME makes the chat `Active`; a successful END ends it. There is one special case: if RESUME returns 19255 from the retained Query node, that response proves the chat is already live there. The SDK therefore marks the chat `Active` with that node as its owner, while still returning the 19255 error to the caller.
+
+Any other server error from RESUME or END leaves the chat `Indeterminate`. In particular, 19236 from END does not prove that PAUSE succeeded; it only says that the retained Query node does not currently have a live chat for that id.
+
+RESUME and END are explicit recovery actions, not status checks. The SDK must not automatically probe other nodes, retry the operation, or try to reconcile the server state.
+
+If the response to RESUME or END is itself lost, the SDK keeps the previous `ChatState` and returns the normal transport error. The [non-replay rule](#retries) still applies.
+
+If the response to BEGIN is lost and the SDK never receives the chat id, it cannot return a usable `Chat`.
+
+## Chat Handle and Re-entrancy
+
+`ChatHandle` contains the SDK state needed to continue a chat later. It is opaque to the application, contains no credentials, and can be serialized so the chat can be continued from another process or SDK instance.
+
+`chat(handle)` restores a `Chat` from that saved state. It is a local operation: it does not contact Query and must not issue RESUME.
+
+The handle must retain:
+
+* chat identity;
+* the Couchbase cluster's `clusterUUID`;
+* the Query node needed to route a live chat;
+* the chat's Knowledge choice;
+* the last-known `ChatState`;
+* for an `Indeterminate` chat, the retained PAUSE target and ambiguity marker.
+
+A handle can only be used with the cluster that created it. If the attaching `Cluster` already knows its `clusterUUID`, the SDK must reject a handle from a different cluster immediately. If the UUID is not yet known, the SDK may defer the check until the first operation that contacts the server. It must validate the UUID before sending that operation, but `chat(handle)` itself must not perform server I/O just to obtain it.
+
+Serialized handles are snapshots; applications should save the latest handle after lifecycle changes.
+The SDK must not repair stale handles by probing Query or interpreting 19236 as a state transition.
+
+Each SDK must support JSON serialization and reconstruction of the complete, opaque `ChatHandle`
+without requiring application access to its internal fields.
+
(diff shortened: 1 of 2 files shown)


rfc/0086-conversational-query.md
```diff
@@ -0,0 +1,714 @@
+# Meta
+
+* RFC Name: Conversational Query
+* RFC ID: 86
+* Start Date: 2026-08-31
+* Owner: Jared Casey
+* Current Status: DRAFT
+* Current Editor: Anirudh Lakhotia
+* Revision: 3
+* Supporting Material: [sdk-design](https://github.com/couchbaselabs/sdk-design/tree/main/server-aligned/totoro/conversational-query)
```
programmatix, line 10: Appreciate you linking to this, but this URL 404s.  Should be https://github.com/couchbaselabs/sdk-design/tree/main/server-aligned/totoro/query-begin-chat

rfc/0086-conversational-query.md
```diff
@@ -0,0 +1,714 @@
...
+new SDK instance.
+
+# General Design
+
+Conversational Query uses the Query service to generate SQL++ from a natural-language prompt and,
+when requested, execute the generated statement under the caller's Couchbase identity. Model access
+uses separate model credentials.
+
+A live chat is held in memory by one Query node. The SDK retains its logical owner for routing and can serialize
+the required client state in a `ChatHandle`. `chat(handle)` reconstructs a `Chat` from a previously
```
programmatix, line 32: Can you check with query team if they have any thoughts/preferences on "Chat" vs "Conversation"?  Maybe feels odd to have Conversational Query not producing Conversations.  I may be overthinking it.

rfc/0086-conversational-query.md
```diff
@@ -0,0 +1,714 @@
...
+## Public API
+
+```
+Keyspaces = list<string>  // SQL++ keyspace paths
+
+Cluster.conversationalQuery([credential ModelCredential], [model ModelOptions])
+    -> ConversationalQuery
+
+ConversationalQuery
+    query(prompt string,
```
programmatix, line 49: I recall discovering on the server-side this was a no-op (contrary to the MB), in that it works like a normal query and doesn't add to the chat at all.  I confirmed that with query team in the chat.  Is that your understanding too?  If so I feel we should drop this `query` as it's actively confusing to users.

rfc/0086-conversational-query.md
```diff
@@ -0,0 +1,714 @@
...
+          [options QueryOptions],
+          [credential ModelCredential])
+        -> ConversationalQueryResult
+
+    beginChat(keyspaces Keyspaces,
+              [options BeginChatOptions],
+              [credential ModelCredential])
+        -> Chat
+
+    chat(handle ChatHandle,
```
programmatix, line 60: Not sure about this naming - it's "verby" indicating I'm starting a chat, and that confuses with beginChat.  

But alternatives are tricky: "resumeChat" confuses with the server feature.  "rehydrateChat" or "deserializeChat" aren't good. 

Dunno how to fix.

rfc/0086-conversational-query.md
```diff
@@ -0,0 +1,714 @@
...
+Chat discovery, model-provider discovery, typed EXPLAIN/ADVISE actions, a public chat-id accessor,
+single-entry Knowledge lookup, and changes to the inactivity timeout after BEGIN are outside this
+revision.
+
+## Public API
+
+```
+Keyspaces = list<string>  // SQL++ keyspace paths
+
+Cluster.conversationalQuery([credential ModelCredential], [model ModelOptions])
```
programmatix, line 45: This doesn't follow existing SDK patterns.  We have calls like `cluster.users()` that return stateless managers you can do calls from, and direct network calls `cluster.query(...)`.  This is new - a stateful parameterised manager. 

Maybe it's ok - we are working with new concepts.  It just feels a little odd at first blush.  Might just need time to bed in.

I think you're mainly doing it this way so the user can optionally fetch and provide the credentials just once, rather than on each beginChat, right?  But there is the credential store.

It also seems odd I can get the stateless KnowledgeManager off this stateful class.

WDYT?

rfc/0086-conversational-query.md
```diff
@@ -0,0 +1,714 @@
...
+    beginChat(keyspaces Keyspaces,
+              [options BeginChatOptions],
+              [credential ModelCredential])
+        -> Chat
+
+    chat(handle ChatHandle,
+         [credential ModelCredential])
+        -> Chat                       // local operation; performs no server I/O
+
+    knowledge() -> KnowledgeManager
```
programmatix: So the knowledge manager was WIP when I looked at the feature.  Could you quickly outline somewhere what it does, or point me at server docs if they're present?
The sdk-design repo would be a fine place for it?

rfc/0086-conversational-query.md
```diff
@@ -0,0 +1,714 @@
...
+    pause([options PauseChatOptions])          -> ConversationalQueryMetaData
+    resume()                                   -> ConversationalQueryMetaData
+    end()                                      -> ConversationalQueryMetaData
+
+    handle() -> ChatHandle
+    state()  -> ChatState
+
+ModelCredential
+    capella(email string, password string, organizationId string)
+    stored(credentialName string)
```
programmatix, line 89: "stored" is a touch vague..  `credentialStore(credentialName)` ?

rfc/0086-conversational-query.md
```diff
@@ -0,0 +1,714 @@
...
+The three credential forms are mutually exclusive. The SDK must not:
+
+* combine Capella credentials with `stored` or `apiKey`;
+* send both `cred_id` and `api_key`; or
+* fall back to another credential form after a failure.
+
+These rules apply to the assembled request because Query silently applies credential precedence;
+Capella credentials cause it to ignore `natural_config`.
+
+If no model credential remains after applying defaults and overrides, Query uses the direct-provider
```
programmatix, line 167: For my own edification - what is the direct provider path?

rfc/0086-conversational-query.md
```diff
@@ -0,0 +1,714 @@
...
+    provider string
+    name string
+    endpoint string
+    region string
+    outputTokenLimit int
+    moderation optional<bool>
+    raw map<string, JsonValue>
+}
+
+QueryOptions {
```
programmatix, line 102: We already have a QueryOptions class.  We want a new one here (copying all of the existing QueryOptions params in), so we don't add a bunch of chat-related stuff onto that path.

rfc/0086-conversational-query.md
```diff
@@ -0,0 +1,714 @@
...
+  Query rejects unsupported provider names.
+* `name`: model identifier. Query passes it verbatim to the selected provider without validating
+  that the model exists.
+* `endpoint`: complete completions URL for the direct-provider path. Query appends no suffix. An
+  explicit endpoint currently permits configurations without a model credential.
+* `region`: provider region; currently used by the Bedrock path.
+* `outputTokenLimit`: output-token limit for the direct-provider path.
+* `moderation`: optional boolean. Unset omits the field and uses Query's default moderation behavior.
+  Explicit `true` or `false` preserves the caller's choice.
+* `raw`: additional top-level `natural_config` entries; values may be any JSON value.
```
programmatix, line 201: Yeah, good call.

rfc/0086-conversational-query.md
```diff
@@ -0,0 +1,714 @@
...
+
+A `ModelOptions` supplied to `conversationalQuery` is the default model configuration. A `Chat`
+created by `beginChat` uses those defaults. A `Chat` reconstructed with `chat(handle)` uses the model
+defaults of the attaching `ConversationalQuery`; the handle contains no model configuration.
```
programmatix, line 205: "A `Chat` reconstructed with `chat(handle)` uses the model defaults of the attaching `ConversationalQuery"

I don't understand why, and it's going to be very confusing to users if they serialize a chat in SDKA using CredsA and ModelParamsA, and resume in SDKB and now they're unexpectedly using CredsB and ModelParamsB.  Definitely violates least surprise.

Why do we need the model params at the reconstruct point?

rfc/0086-conversational-query.md
```diff
@@ -0,0 +1,714 @@
...
+fields are not merged. An explicitly empty replacement inherits no model fields or raw entries.
+
+Credential selection is independent of model replacement.
+
+### Raw model options
+
+`raw` exists only as a forward-compatibility mechanism for future Query fields. Current Query
+versions silently discard unrecognized `natural_config` keys; `raw` is not provider passthrough.
+
+The SDK must reject these keys in `raw` because they already have typed representations:
```
programmatix, line 217: I'd maybe be careful of this, and perhaps just set a precedence rule instead.
In future when NewModelFieldX is added, SDKs will add it and could carelessly add to this list.  That immediately breaks any existing apps that have set NewModelFieldX via raw.

rfc/0086-conversational-query.md
```diff
@@ -0,0 +1,714 @@
...
+
+`provider`, `model`, `endpoint`, `region`, `output_token_limit`, `moderation`, `cred_id`, `api_key`.
+
+This applies even when the corresponding typed option or credential is unset. Allowing the same
+setting through two paths would make precedence SDK-specific.
+
+Other `raw` entries are added unchanged to `natural_config`. They cannot override Query-controlled
+settings such as temperature or seed.
+
+The SDK must reject options with no wire representation for the selected credential path; see
```
programmatix, line 227: Sorry, what does this mean; rejecting unknown `raw` fields doesn't make sense so I must be misunderstanding.

rfc/0086-conversational-query.md
```diff
@@ -0,0 +1,714 @@
...
+
+Execution is enabled by default. `execute(false)` requests SQL++ generation without execution.
+
+## Chat Lifecycle
+
+A live chat is held in memory by one Query node. It can disappear if that node restarts, is failed over or rebalanced out, or if the chat is evicted or expires from inactivity. Unless the chat has been paused first, that state is not persisted elsewhere.
+
+A paused chat is stored persistently by Query and has its own server-controlled expiry. SDKs must not try to predict when live or paused chats expire or reject requests based on hard-coded server limits.
+
+Applications must not perform multiple operations on the same chat concurrently. SDKs are not required to coordinate concurrent use of separate `Chat` instances or copies of the same `ChatHandle`.
```
programmatix, line 245: "Applications must not perform multiple operations on the same chat concurrently."
OOI - what's the issue if that happens?  I don't think there's any threading issues on the client side - does server have a bug here?

rfc/0086-conversational-query.md
```diff
@@ -0,0 +1,714 @@
...
+
+### State
+
+`Chat.state()` reports the state the SDK last observed for the chat. It does not query the server or guarantee that the chat is still in that state.
+
+| State | Meaning |
+| --- | --- |
+| `Active` | The SDK last saw the chat as live and knows which Query node owns it. |
+| `Paused` | PAUSE succeeded, so the SDK no longer treats any Query node as the live owner. |
+| `Indeterminate` | The SDK sent PAUSE but lost the response, so it does not know whether the chat is still live or was successfully paused. |
```
programmatix, line 255: Nb elsewhere in SDKs we've used "ambiguous", maybe we should stick to that convention here.
We could also be less, uh, ambiguous and use `AmbiguouslyPaused`?  I may be overthinking.

rfc/0086-conversational-query.md
```diff
@@ -0,0 +1,714 @@
...
+
+A live chat is held in memory by one Query node. It can disappear if that node restarts, is failed over or rebalanced out, or if the chat is evicted or expires from inactivity. Unless the chat has been paused first, that state is not persisted elsewhere.
+
+A paused chat is stored persistently by Query and has its own server-controlled expiry. SDKs must not try to predict when live or paused chats expire or reject requests based on hard-coded server limits.
+
+Applications must not perform multiple operations on the same chat concurrently. SDKs are not required to coordinate concurrent use of separate `Chat` instances or copies of the same `ChatHandle`.
+
+### State
+
+`Chat.state()` reports the state the SDK last observed for the chat. It does not query the server or guarantee that the chat is still in that state.
```
programmatix, line 249: Per previous discussion, I'm wary of cached mutable state in general.
Remind me why this is desired?
Is there some naming that would better indicate that this is a cached state from our last network call?

rfc/0086-conversational-query.md
```diff
@@ -0,0 +1,714 @@
...
+
+| State | Meaning |
+| --- | --- |
+| `Active` | The SDK last saw the chat as live and knows which Query node owns it. |
+| `Paused` | PAUSE succeeded, so the SDK no longer treats any Query node as the live owner. |
+| `Indeterminate` | The SDK sent PAUSE but lost the response, so it does not know whether the chat is still live or was successfully paused. |
+
+`Indeterminate` prevents another `ask` because PAUSE may already have removed the live chat from its owner.
```
programmatix, line 257: This half-suggests it's a rule we should apply client-side, which I assume you don't mean.

rfc/0086-conversational-query.md
```diff
@@ -0,0 +1,714 @@
...
+
+| Operation | Permitted local state | Effect of a successful response |
+| --- | --- | --- |
+| `ask` | Active | Stay Active. |
+| `pause` | Active | Become Paused. |
+| `resume` | Paused or Indeterminate | Become Active; the Query node that serves RESUME becomes the new owner. |
```
programmatix, line 263: Ah you did mean it's enforced client-side.
I would suggest not.
Perhaps a chat gets itself into an odd state due to a bug, and someone runs RESUME CHAT manually at the workbench.  We're now out of sync with the real server state - this is why I don't love caching and making decisions from mutable state.

I would say just send it through to the server, and handle/surface the resultant errors if e.g. user RESUMEs a non-PAUSED chat.

rfc/0086-conversational-query.md
```diff
@@ -0,0 +1,714 @@
...
+ConversationalQuery
+    query(prompt string,
+          keyspaces Keyspaces,
+          [options QueryOptions],
+          [credential ModelCredential])
+        -> ConversationalQueryResult
+
+    beginChat(keyspaces Keyspaces,
+              [options BeginChatOptions],
+              [credential ModelCredential])
```
programmatix, line 57: Double-optional is a bit of an odd pattern.  Can credential just become an option under BeginChatOptions.

(Assuming credentials truly are optional - I had a question about that further down)

rfc/0086-conversational-query.md
```diff
@@ -0,0 +1,714 @@
...
+
+After a successful PAUSE, the SDK marks the chat Paused and no longer keeps a live owner for it.
+
+`summarize` controls whether Query may summarize the conversation before saving it:
+
+- unset: let Query decide;
+- `true`: request summarization;
+- `false`: do not summarize.
+
+If summarization is possible, PAUSE may invoke the model, so the SDK sends the chat's model configuration and available model credential. When `summarize` is `false`, it sends neither. Because PAUSE may include a model call, its default request timeout must not be shorter than the SDK's model-turn timeout.
```
programmatix, line 302: Should model config and credentials be optional here too then?

rfc/0086-conversational-query.md
```diff
@@ -0,0 +1,714 @@
...
+
+A handle can only be used with the cluster that created it. If the attaching `Cluster` already knows its `clusterUUID`, the SDK must reject a handle from a different cluster immediately. If the UUID is not yet known, the SDK may defer the check until the first operation that contacts the server. It must validate the UUID before sending that operation, but `chat(handle)` itself must not perform server I/O just to obtain it.
+
+Serialized handles are snapshots; applications should save the latest handle after lifecycle changes.
+The SDK must not repair stale handles by probing Query or interpreting 19236 as a state transition.
+
+Each SDK must support JSON serialization and reconstruction of the complete, opaque `ChatHandle`
+without requiring application access to its internal fields.
+
+`ChatHandle` must be portable across processes and SDK instances. Its serialized format is SDK-specific and may evolve between versions, so implementations should be able to recognize unsupported handle formats.
```
programmatix, line 374: Where is the serialized format defined?

rfc/0086-conversational-query.md
```diff
@@ -0,0 +1,714 @@
...
+
+# Implementation Details
+
+## Node Affinity
+
+A live chat is held in memory by one Query node. The node that serves BEGIN or RESUME becomes the chat's owner. Query does not route requests to the owner by chat id, so the SDK must send ASK, PAUSE, and END back to that node.
+
+The SDK stores the owner's logical node identity, not the Query endpoint used for the original request. The logical identity is Query's canonical node name: the canonical hostname plus the default-network management port.
+
+Before each affine operation, resolve the retained owner using normal SDK topology handling.
```
programmatix: affine?

rfc/0086-conversational-query.md
```diff
@@ -0,0 +1,714 @@
...
+
+If the owner is absent or has no usable Query endpoint, return a routing error. The SDK must not
```
programmatix, line 406: "routing error" - please be explicit on which SDK3 error is raised.
This is a rare case that the user might want to handle programmatically with a retry, so could be worth a new exception.

rfc/0086-conversational-query.md
```diff
@@ -0,0 +1,714 @@
...
+#### Name and keyspace encoding
+
+`name` is supplied by the application as a plain string. The SDK encodes the entire value as one SQL++ identifier and doubles any embedded backticks. Applications must not pre-quote it.
+
+`keyspace` is either:
+
+* a bucket name, such as `travel-sample`; or
+* a three-part `bucket.scope.collection` path, such as `travel-sample.inventory.hotel`.
+
+The SDK must reject empty components, any other number of components, namespace prefixes, and pre-quoted identifiers.
```
programmatix, line 499: Discarding pre-quoted identifiers seems harsh, since that's the user performing a practice they may be used to from the query workbench.

rfc/0086-conversational-query.md
```diff
@@ -0,0 +1,714 @@
...
+#### Knowledge text
+
+The SDK must bind Knowledge text as the named parameter `$value`.
+
+#### Listing entries
+
+`getAll()` uses the projection above, escaping the reserved catalog field names. Construct
+`KnowledgeEntry.keyspace` by joining `bucket`, `scope`, and `collection` with dots. A bucket-only
+input such as `travel-sample` is therefore returned as `travel-sample._default._default`; both
+forms identify the same entry.
```
programmatix, line 522: No quoting needed?

rfc/0086-conversational-query.md
```diff
@@ -0,0 +1,714 @@
...
+
+`create` uses `CREATE KNOWLEDGE` without `IF NOT EXISTS`. If the entry already exists, Query returns error 20105.
+
+`drop` uses `DROP KNOWLEDGE` without `IF EXISTS`. If the entry does not exist, Query returns error 20103.
+
+Other server-supported Knowledge grammar is outside this SDK API.
+
+## Results
+
+`ConversationalQueryResult` extends normal Query result handling with the metadata below.
```
programmatix, line 548: Nb "extends" suggests "inherit".  I would suggest making it explicit that it's a copy, allowing the two interfaces to go their own separate ways in future.

rfc/0086-conversational-query.md
```diff
@@ -0,0 +1,714 @@
...
+### Token usage
+
+Conversational Query can return:
+
+```text
+requestTokens() -> optional<TokenUsage>
+chatTokens()    -> optional<TokenUsage>
+
+TokenUsage {
+    promptTokens int
```
programmatix, line 577: I'd suggest making these all optional.  Looks like the sort of metadata that may change in future as LLMs evolve.

rfc/0086-conversational-query.md
```diff
@@ -0,0 +1,714 @@
...
+
+`requestTokens` reports token usage for the current request.
+
+`chatTokens` reports cumulative token usage for the chat and, when present, must also be exposed in metadata returned from PAUSE, RESUME, and END.
+
+Both fields are optional. Query may omit an all-zero token object, and the Capella path does not report token usage. An absent value therefore means only that token usage was not reported; the SDK must not replace it with an all-zero `TokenUsage`.
+
+The SDK must use the `totalTokens` value returned by Query. It must not calculate it from `promptTokens` and `completionTokens` or reject the response if those values do not add up.
+
+`naturalLanguageProcessingTime` remains part of the normal Query metrics object and may be exposed through the SDK's existing Query metrics type.
```
programmatix, line 594: `QueryMetrics` ?  Hmm.  I could see us slowly 'contaminating' the regular query objects with a bunch of CQ-specifics.  I think we need to copy all classes involved, not share between regular query and ask.

rfc/0086-conversational-query.md
```diff
@@ -0,0 +1,714 @@
...
+
+Except for the rules below, normal Query error handling (RFC 0058) applies.
+
+Preserve the server code, message, and nested `reason` in the Query error context, subject to
+credential redaction. Some Conversational Query errors carry useful detail only in `reason`;
+SDKs must not rely on `msg` alone or rewrite server error text.
+
+| Code  | SDK handling                                                                                                                                     |
+| ----- | ------------------------------------------------------------------------------------------------------------------------------------------------ |
+| 19220 | `RateLimited` (RFC 0058, #21)                                                                                                                    |
```
programmatix, line 616: Nb it would be useful to have a little more context on these errors.  Maybe copy from the query docs if available?

rfc/0086-conversational-query.md
```diff
@@ -0,0 +1,714 @@
...
+A request may already have called the model, changed chat history, or executed generated SQL++ before an error is returned. Replaying it could therefore repeat work or change state twice. This applies even when `execute(false)` is used or the generated Query is marked `readonly`.
+
+The SDK may use normal retry handling (RFC 0049) only when the request is known not to have reached Query.
+For a chat operation, any such retry must target the same Query node or retained PAUSE target.
+
+Query performs its own retries for model and generation work. SDKs must not add another generic retry loop for model-provider failures.
+
+Conversational Query-specific 192xx errors omit Query's `retry` flag; ordinary Query errors from
+executing generated SQL++ may include it. The SDK should preserve and handle the flag under normal
+Query error rules, but it must never override the Conversational Query non-replay rule.
```
programmatix, line 646: I'm confused by this wording.  Should we respect the flag or not?

rfc/0086-conversational-query.md
```diff
@@ -0,0 +1,714 @@
...
+Because the feature-control check happens before chat lookup, a 19222 response also does not tell the SDK whether the chat exists.
+
+# Documentation
+
+SDK documentation should cover one-shot queries, chat lifecycle and persistence, `ChatHandle` usage, Knowledge, and generation without execution. It should also describe the user-visible consequences of Indeterminate PAUSE outcomes, stale handles, chat expiry, Knowledge prompt reuse, and serialized access to a chat.
+
+Model-credential documentation must describe the prerequisites for each credential form. In particular:
+
+* `capella(email, password, organizationId)` uses a Capella console login rather than an organization API key, requires accepted iQ terms, and is not available to accounts that can authenticate only through SSO or social login.
+* `stored(credentialName)` requires node-to-node encryption on all nodes or `n2nEncryptionOverride=true` in `/settings/credentialStore`.
```
programmatix, line 673: Nb this feels somewhat non-DRYed and the sort of info that goes stale.  I'd suggest linking to relevant part of the server docs instead.

rfc/0086-conversational-query.md
```diff
@@ -0,0 +1,714 @@
...
+
+Operator documentation should reference server configuration for the credential store, chat persistence, model connectivity, and resource limits. It should also note that summarization can add model latency and token cost, and that chat operations require the SDK to identify the serving Query node.
+
+Data-handling documentation must describe the Couchbase authorization boundary and the information that may be sent to model providers. Chat context can include prompts, generated statements and Query correction rounds, but not result rows. The `slm` provider can additionally include representative sample values. Knowledge belongs to its target bucket and is not currently covered by backup and restore ([MB-73627](https://jira.issues.couchbase.com/browse/MB-73627)).
+
+# Changelog
+
+* Revision 3 - 2026-09-10 (by Anirudh Lakhotia)
```
programmatix, line 681: Nb we only start doing revisions after releasing the thing - no need to do it during early design work (you'll drive yourself nuts!)

rfc/0086-conversational-query.md
```diff
@@ -0,0 +1,714 @@
...
+
+  * Expanded the original draft with chat lifecycle, credential types, and result handling.
+
+* Revision 1 - 2026-08-31 (by Jared Casey)
+
+  * Original draft.
+
+# Signoff
+
+Before signoff, each SDK team must confirm that its Query dispatch path can identify the specific node
```
programmatix, line 698: This last para is a bit odd.

rfc/0086-conversational-query.md
anirudhlakhotia: Updated

rfc/0086-conversational-query.md
anirudhlakhotia: This is a different case. Ordinary SQL++ with a chat id ignores the id

`query()` is a one-shot `USING AI` request with no chat state

rfc/0086-conversational-query.md
anirudhlakhotia: The direct-provider path is where Query sends model requests directly to the configured model provider rather than through the Capella iQ intermediary

I'll add a small definition to the RFC

rfc/0086-conversational-query.md
anirudhlakhotia: I confirmed this with Query - overlapping requests for a chat are serviced one at a time, but Query does not guarantee their relative order.

The RFC now requires the SDK to support concurrent operations safely and states only that no ordering guarantee is provided.

rfc/0086-conversational-query.md
anirudhlakhotia: Ended up removing the public lifecycle state entirely, so there is no longer a public `Indeterminate`/ `Ambiguous` state to name

rfc/0086-conversational-query.md
anirudhlakhotia: As per our discussion on slack, we don't need to maintain any state on the client-side
The SDK only retains the routing and recovery information that it will actually need internally

rfc/0086-conversational-query.md
anirudhlakhotia: Yes. Model config and credentials are optional here too.

If PAUSE may summarize, the SDK sends the effective model config and credential. With `summarize(false)`, no model inputs are sent and any model/credential overrides are ignored

rfc/0086-conversational-query.md
programmatix: Ahh, I see.  Ok, fair enough.

rfc/0086-conversational-query.md
programmatix: Ok and that is pre-configured at the server or UI level?

rfc/0086-conversational-query.md
```diff
@@ -0,0 +1,653 @@
...
+    beginChat(keyspaces Keyspaces,
+              [options BeginChatOptions])
+        -> Chat
+
+    attachChat(handle ChatHandle,
+               [options AttachChatOptions])
+        -> Chat                       // local operation; performs no server I/O
+
+KnowledgeManager
+    upsert(name string, keyspace string, value string)   // CREATE OR REPLACE
```
programmatix, line 54: keyspace, name, value ordering would feel more natural perhaps?  As it goes from least to most specific, and groups name & value.
I know it's this way round in the SQL, but in an SDK it looks odd.

rfc/0086-conversational-query.md
```diff
@@ -0,0 +1,653 @@
...
+    clientContextId() -> string
+    status() -> QueryStatus
+    warnings() -> list<QueryWarning>
+    metrics() -> optional<QueryMetrics>
+    requestTokens() -> optional<TokenUsage>
+    chatTokens() -> optional<TokenUsage>
+    naturalLanguageProcessingTime() -> optional<Duration>
+
+TokenUsage {
+    promptTokens int
```
programmatix: Nb still feel we may be better on making all fields in TokenUsage optional.  Feels a very "whatever we got from a particular model" sort of JSON blob.

rfc/0086-conversational-query.md
```diff
@@ -0,0 +1,653 @@
...
+    execute bool
+    output ConversationalQueryOutput
+    knowledge bool
+}
+
+BeginChatOptions {
+    model ModelOptions
+    credential ModelCredential
+    inactivityTimeout Duration
+    knowledge bool
```
programmatix, line 122: Would this be clearer as useKnowledge?

rfc/0086-conversational-query.md
```diff
@@ -0,0 +1,653 @@
...
+
+### Ordinary Query controls
+
+The listed Query controls retain their existing types, semantics, and wire mappings. The SDK must send any controls the caller explicitly sets.
+
+`timeout` uses the existing Query timeout handling. A timeout reported by the SDK does not guarantee that Query has stopped processing the request.
+
+BEGIN uses `BeginChatOptions.timeout` when set, and PAUSE uses `PauseChatOptions.timeout` when set. Otherwise, lifecycle operations use the cluster's default Query timeout.
+
+`rawParameters` follows RFC 0056. The SDK must reject any raw parameter whose name, compared case-insensitively, begins with `natural`.
```
programmatix, line 191: Why this rule?

rfc/0086-conversational-query.md
```diff
@@ -0,0 +1,653 @@
...
+
+The SDK does not expose prepared statements, named or positional parameters, `readonly`, `preserveExpiry`, or `queryContext` for Conversational Query, and must send these requests as adhoc.
```
programmatix, line 193: `queryContext`
Hmm - this brings up an important problem of cluster vs scope level queries.  Needs more consideration here.

`adhoc` makes sense to exclude.

Why are the others not exposed?  Named & positional params are usually particularly crucial for security reasons - are they not combinable with USING AI?  Can I not do "USING AI give me all the orders from customer $namedCustomer"? 

`readonly` seems particularly important for (I'm assuming query supports this) preventing AI-generated SQL from doing writes.  Though actually I half-recall they only auto-execute SELECTs anyway - in which case, fair enough.

rfc/0086-conversational-query.md
```diff
@@ -0,0 +1,653 @@
...
+| --- | --- |
+| `capella(email, password, organizationId)` | `natural_cred` = `email:password`; `natural_orgid` = organization id |
+| `credentialStore(credentialName)` | `natural_config.cred_id` |
+| `apiKey(key)` | `natural_config.api_key` |
+
+The SDK sends the effective model credential on every request that may invoke the model. BEGIN, RESUME, and END carry no model credential.
+
+The SDK must not fall back to another credential form after a failure.
+
+If no model credential is set after resolving defaults and overrides, the request uses the direct-provider path.
```
programmatix, line 211: This remains a little ambiguous on whether this action is expected to be on the client or server side.

rfc/0086-conversational-query.md
```diff
@@ -0,0 +1,653 @@
...
+
+KnowledgeEntry {
+    name string
+    keyspace string
+    value string
+}
+
+Chat
+    ask(prompt string, [options AskOptions])   -> ConversationalQueryResult
+    pause([options PauseChatOptions])          -> ConversationalQueryMetaData
```
programmatix, line 67: Nb SDK3 convention is that options blocks are just MethodNameOptions, e.g. PauseOptions here.

rfc/0086-conversational-query.md
```diff
@@ -0,0 +1,653 @@
...
+
+The SDK must not issue RESUME merely to probe chat state. A successful RESUME consumes the persisted chat and restores it as a live chat.
+
+### End
+
+`end` ends a live chat and releases its live slot. The SDK sends END to the retained owner.
+
+END does not delete a paused chat. If no owner is retained, `end()` fails locally with `InvalidArgumentException` (RFC 0058, #3); the application must RESUME before END can be sent.
+
+After END succeeds, the local `Chat` wrapper is closed. The SDK rejects further server operations, and `handle()` fails locally on that object.
```
programmatix, line 353: >The SDK rejects further server operations

What error here?

rfc/0086-conversational-query.md
```diff
@@ -0,0 +1,653 @@
...
+
+Previously serialized handles cannot be revoked by closing this local `Chat`.
+
+## Chat Handle
+
+`ChatHandle` is an opaque, serializable representation of the client state needed to continue and route a chat. It does not represent server lifecycle state.
+
+`attachChat(handle)` is local and must not issue RESUME. The returned `Chat` uses the attaching `ConversationalQuery` defaults unless `AttachChatOptions` replaces them. The handle contributes no model or credential defaults.
```
programmatix, line 361: A very tricky thing to name and I don't hate attachChat

rfc/0086-conversational-query.md
```diff
@@ -0,0 +1,653 @@
...
+
+The handle must retain:
```
programmatix, line 363: Can we get a solid JSON format here - exact field names

rfc/0086-conversational-query.md
```diff
@@ -0,0 +1,653 @@
...
+
+* the chat id;
+* the Couchbase cluster's `clusterUUID`;
+* the retained owner identity, or the absence of one;
+* the chat's Knowledge choice.
+
+A handle can be used only with its originating cluster. If the attaching `Cluster` already knows its `clusterUUID`, the SDK must reject a foreign handle immediately. Otherwise, it must validate the UUID before the first server operation.
```
programmatix, line 370: This may be overkill that may very occasionally backfire - I think the cluster UUID can change in some situations.

rfc/0086-conversational-query.md
```diff
@@ -0,0 +1,653 @@
...
+
+`attachChat(handle)` must not perform I/O solely to obtain the UUID.
+
+A serialized handle captures routing state that may later become stale. The SDK must not probe Query to validate or refresh it.
+
+Each SDK must support serializing `ChatHandle` to JSON and reconstructing it from that JSON. The format is SDK-specific and must be versioned so incompatible handles can be rejected.
```
programmatix, line 376: What is the need for it to be SDK-specific?
Versioning is a solid idea, please detail.

rfc/0086-conversational-query.md
```diff
@@ -0,0 +1,653 @@
...
+
+A handle must not contain Couchbase or model credentials, model configuration, a resolved Query endpoint, turn-specific hints, keyspaces, or the chat inactivity timeout.
+
+## Knowledge
+
+`KnowledgeManager` is obtained from `Cluster` and uses the cluster's Couchbase identity. Knowledge operations do not use model configuration, model credentials, or chat routing.
```
programmatix, line 382: I think everything in KnowledgeManager be accessed via SQL statements, and I'm tempted to suggest we remove it from the SDK to reduce scope.  Am open to pushback on that though.

rfc/0086-conversational-query.md
```diff
@@ -0,0 +1,653 @@
...
+
+# Implementation Details
+
+## Node Affinity
+
+A live chat is held in memory by one Query node, and Query does not route a chat request to its owner by chat id. The SDK must therefore send ASK, PAUSE, and END to the retained owner.
+
+The SDK retains the owner's logical node identity, not a resolved Query endpoint. The identity is Query's canonical node name: the canonical hostname plus the non-TLS management port (`host:8091` by default), as reported by the `node` field of `system:natural_chats`, even when the SDK connects over TLS.
+
+Before sending a chat operation to the retained owner, the SDK resolves that logical identity through normal topology handling for the selected network.
```
programmatix, line 398: Hmm.  This feels like a little world of landmines.  There are situations (I'm thinking of private links which uses a load balancer) where the hostname is always the same and the port identifies both the real node & the service.  I don't think that's compatible with query's node name concept.

rfc/0086-conversational-query.md
anirudhlakhotia: Yeah, I agree. Knowledge management feels like a separate API surface from CQ, especially since it can already be handled through ordinary Query. I'm not sure we even need first-class SDK support for it right now, so I'll keep this RFC focused on consuming Knowledge

If that changes later, I guess it can just be a separate RFC

rfc/0086-conversational-query.md
anirudhlakhotia: Yeah, I need to think about this a bit more 🤔 

Would really appreciate any pointers on how we usually handle node identity in cases like Private Link or alternate addresses. I'll dig into the topology handling on my side too.

programmatix: Sorry, I missed this.  Did the Slack conversation address all questions?

rfc/0086-conversational-query.md
programmatix: Yep, seems very reasonable.

PR #166, opened 2026-09-16
RFC-52: Support for FTS queries with user-defined score/filter functions

DemetrisChr:
https://jira.issues.couchbase.com/browse/SDK-44

Adding `CustomScoreQuery` and `CustomFilterQuery` search query (non-vector) types. These are can be used in the same way as any other `SearchQuery` implementation.

Accompanied by FIT tests: https://review.couchbase.org/c/transactions-fit-performer/+/253428

[MB-73982]: https://couchbasecloud.atlassian.net/browse/MB-73982?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

This feature is disabled on the cluster by default. Users have to enable it via the UI or the REST API:
```
curl -X PUT -u Administrator:password -H "Content-Type: application/json" -d '{"customScriptQueriesEnabled": "true"}' http://192.168.106.128:8094/api/managerOptions
```


dnault (approved):

rfc/0052-sdk3-full-text-search.md
```diff
@@ -587,6 +587,76 @@ A query that matches nothing.
...
+      String source
+  );
+  ...
+}
+```
+
+```java
+class CustomScoreQuery implements SearchQuery {
+  CustomScoreQuery fields(String... fields);
+  CustomScoreQuery parameters(JSONObject p);
```
programmatix, line 621: It might be worth clarifying (since JsonObject is a rather Java SDK specific construct) that JSON is exposed however it is in the rest of the SDK, and should be able to support arbitrary nested JSON.

If the SDK does not have a good native way to express JSON then a suitable method is to support native collections - maps & lists.  (That's what we landed on in the context of, IIRC, Protostellar)

rfc/0052-sdk3-full-text-search.md
```diff
@@ -587,6 +587,76 @@ A query that matches nothing.
...
+class CustomFilterQuery implements SearchQuery {
+  CustomFilterQuery fields(String... fields);
+  CustomFilterQuery parameters(JSONObject p);
+}
+```
+
+#### Compatibility
+
+Awaiting cluster cap ([MB-73982](https://jira.issues.couchbase.com/browse/MB-73982)).
+
```
programmatix, line 695: I think we should include (possibly on a ticket rather than the RFC, up to you), that the documentation should highlight some quirks the user needs to know:

* A mention that custom scoring should not be used together with disableScoring.  (Source https://couchbase.slack.com/archives/C0BU965HJNT/p1788858007574109?thread_ts=1788433533.573109&cid=C0BU965HJNT answer 2)
* That the user probably wants to specify some fields - I see it as only technically optional to omit them, since then your function receives no data from the doc at all.
* Possibly that it is EE only.
* Possibly that it must be enabled on cluster.

We would ideally test (going a bit above and beyond):

* How/if SDKs surface partial results from FTS - an error mid-stream.  That is going to be much more possible now there's a user-provided chunk of JS logic.

programmatix (commented):
Yeah this looks like where I landed too (https://github.com/couchbaselabs/sdk-design/blob/main/server-aligned/totoro/fts-udf/sdk.md).  Broad strokes LGTM, just a couple of suggestions.

rfc/0052-sdk3-full-text-search.md
DemetrisChr: > That the user probably wants to specify some fields - I see it as only technically optional to omit them, since then your function receives no data from the doc at all.

I think there are valid use cases where no fields are needed (e.g. you want to do some kind of transformation on the score itself irrespective of the rows), but I'll make it clearer what the absence of `fields` means.

> How/if SDKs surface partial results from FTS - an error mid-stream. That is going to be much more possible now there's a user-provided chunk of JS logic.

~According to FTS's design, if there is a runtime error in the Javascript function, for custom_filter queries, the row is dropped from the results. For custom_score queries, the original score of the row is preserved. No information about the error reaches the SDK.~ (Edit: Not seeing this behaviour as described in the design doc,  I'll raise this with the FTS team)

programmatix: > (e.g. you want to do some kind of transformation on the score itself irrespective of the rows)

Unconvinced tbh - if it's a transformation that doesn't need the row data, it would seem much cheaper to do it in the app as the results come in.  I'm struggling to see anything very useful you could do with the metadata alone.  Boost the score based on a key prefix perhaps, though you could probably do that more easily with one of the existing FTS queries + boost.

On errors, Rohit described the expected behaviour on https://couchbase.slack.com/archives/C0BU965HJNT/p1788858007574109?thread_ts=1788433533.573109&cid=C0BU965HJNT, which does seem to differ to the design doc.

rfc/0052-sdk3-full-text-search.md
DemetrisChr: Updated

rfc/0052-sdk3-full-text-search.md
DemetrisChr: I've updated it to mention that you probably want fields to be set, that would be useful for API docs. I agree that most queries will want access to some fields. I don't think that means we need any changes to the API or to make it required, since we still want to have the option of not setting it (which is consistent with the FTS API).

Yeah, that seems to be the error behaviour. I think that doesn't affect our design, it's just like any other error. I'll make sure to add a FIT test covering partial result scenarios to check that SDKs already do the right thing.

programmatix: I agree, we don't want to make API changes for this.  I did mention in the original comment that this fields quirk should be mentioned only in the docs.

And yes, it appears to be standard FTS error behaviour; I just think that as this is the first time SDKs are likely to be encountering FTS partial results in the wild, we might unearth some long-extant bugs - thanks for adding some testing for it.

programmatix (approved):

davidkelly (approved):
👍

dnault (approved):

thejcfactor (approved):
LGTM 👍

